PR #13, opened 2018-10-04
Several additions from my private repos

jaabell:
Merging in stuff I've been writing and playing with for the past year or so. This compiles without warnings on Ubuntu Linux using Intel 18 Update 3 compilers as well as gcc-5.5. Have not tested on Win64. The GmshRecorder is working nicely, but could use some improvements regarding usability.


jaabell:
Fixing merge conflict with base.

PR #58, opened 2018-11-29
Adding a 2D wheel-rail element to simulate the vertical wheel-rail interaction (WRI).

Lew007:
A 2D wheel-rail element is implemented to simulate the vertical wheel-rail interaction (WRI). The time- and location- varying WRI makes the simulation of vehicle-track systems complicated. This new element simulates the WRI using a location dependent internal resisting force, which enables the whole finite element (FE) model to remain unchanged in the analysis. Each WRI element includes a wheel node and all the nodes of rail elements that the wheel may potentially pass over during the movement.
The details can be referred to "Gu Q, Liu Y, Guo W, et al. A Practical Wheel-Rail Interaction Element for Modeling Vehicle-Track-Bridge Systems[J]. International Journal of Structural Stability and Dynamics, 2018: 1950011."

### DIFF
--- a/SRC/element/TclElementCommands.cpp
+++ b/SRC/element/TclElementCommands.cpp
@@ -48,6 +48,7 @@
 #include <MultipleShearSpring.h>
 #include <KikuchiBearing.h>
 #include <YamamotoBiaxialHDR.h>
+#include<WheelRail.h>
 
 extern 
 #ifdef _WIN32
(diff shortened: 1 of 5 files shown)


Lew007:
Hi, Frank, I delete my (old) repository and fork your version again. So there is only one commit here.

fmckenna:
i will move some code from TclElement to the WheelRail element .. is this really an element!

PR #59, opened 2018-12-04
Fix compiling errors in WheelRail

zhuminjie:
Fix errors and add object files in makefile


fmckenna:
there is another similar pull request .. will use it instead

PR #60, opened 2018-12-05
fix compile errors and add travis-ci

frqc:
Fix compilation lead by the 2D wheel-rail element.
Add Makefile.def for travis-ci and its yaml configuration. 

With continuous integration by travis-ci, we can finally say good bye to compilation error brought by new commit by [compiling the new commit](https://docs.travis-ci.com/user/for-beginners/) on a virtual machine automatically. 

I added a new Makefile.def for travis-ci which links the reliability and optimization libs by default. The virtual machine which will run the compilation is Ubuntu 16.04. Other configurations can be seen in .travis.yml. The current configuration successfully built the opensees (but haven't test python for opensees yet). 

If this pr is accepted, to enable auto-compilation the owner of this repo will still have to go to [here ](https://travis-ci.org/)to give access to travis-ci and turn on the auto-compilation switch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ dist: xenial
 
 before_install:
   - sudo apt-get install -y g++ gcc gfortran make python3 python3-dev tcl8.6 tcl8.6-dev
-  - mkdir ../bin
-  - mkdir ../lib
+  - mkdir bin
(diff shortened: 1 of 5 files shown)


frqc:
Compilation for python and a small import test are added

PR #80, opened 2019-02-14
Intel MKL PARDISO - Parallel Direct Sparse Solver Interface

imeg:
https://software.intel.com/en-us/mkl-developer-reference-fortran-intel-mkl-pardiso-parallel-direct-sparse-solver-interface


imeg:
add Intel MKL header library

fmckenna:
we cannot make the intel headers available you have placed in OTHER  .. The copyright on the intel header files is not open source .. you will have to remove the files and assume the user can include the MKL header files in the include paths

PR #97, opened 2019-04-18
adding new materials, recorder types, commands and a few minor improvements

sa-jalali:


fmckenna:
i am not going to accept the changes
1. we don't want the location of where is installed changed
2. ok
3. ok
4. we have the eleResponse and nodeResponse commands that do the same thing. why are you adding this change?
5. needs some more study as to code added.
5

sa-jalali:
Thank you for your response.
Actually, I am not much experienced in github. I would better requested new pools after the first ones were accepted or rejected.
About the changes, I have to say that:
1. Yes, I guessed so because that was a change appropriate for my machine. However, I didn't know (and still don't know actually) how to have the github exclude some changes I make to the code to have it work for me from those being committed to the OpenSees/master.
2. Thanks
3. Thanks.

4. The eleResponse and nodeResponse commands do not work with:
a) Drift Recorder (we need to use for checking maximum inter-story drift to avoid unnecessary attempts on an already-collapsed structure while trying to achieve convergence in a near-to-collapse seismic excitation 
b) Envelope-type recorders (we use "recorderValue" command for collecting the min. and max. responses over a given time interval before the response is read. This is done by "resetting" the envelope-recorder to recalculate the min and max. values during the rest of analysis and before another read using the "recorderValue" command)
In addition to above, this new command has a more general and versatile functionality because leaves response recording to "recorder" objects that may be extended at any time in future.

5. Thank you for your time. To help understanding the changes, I can provided some details about them:
a) changes in uniaxialMaterial classes
- a new virtual method is added to the base class uniaxialMaterial. The method signature is: "virtual double getEnergy(void)" and returns 0 as the default implementation (I have considered code maintenance issues by not making it pure virtual)
- the new method is overridden in 10 of frequently-used uniaxialMaterial's including Bilin, Steel01, Steel02, Concrete01 and others.
- in some cases, the material already computed energy during analysis excursions (like Bilin for which only the .h file was modified to return the energy value by the new method)
- in other cases, a new variable is added which is initiated to 0 in all constructors and is only updated upon committing the material status (just before the old commit values of strain and stress are overwritten with the new ones)
b) changes in section classes (fiberSection2d, fiberSection3d and sectionAggregator):
- the base class is not modified
- a new public method "double getEnergy(void)" is added whose implementation depends on uniaxialMateril's "getEnergy". That is, the function iterates over all fibers, obtains energy from their material, multiplies it by fiber area and adds it to the section's energy.

c) changes to the fiber beamColumn elements (2d and 3d versions of disp- and force-based beamColumn 's)
-the energy-per-length value returned by each section is multiplied by the length of each integration point and is summed up over all points

d) recorder-specific changes
-in all classes, the setResponse and getResponse classes are modified to recognize the "energy" option and call their "getEnergy" method for obtaining the response.
-in case of materials, this is done in the base class "UniaxialMaterial"

Thank you for reaching here :)

sa-jalali:
I would be thankful if you could guide me how I should separate the commits in order to have the pull accepted.
Thanks again,
Alireza

sa-jalali:
Sorry for the mess, but I finally made sure that the project settings were set back to their original values.

sewkokot:
There are two issues with this pull request.

1. I have to change the following line in SRC/tcl/commands.cpp 
from:
#include <Recorder/Recorder.h> //SAJalali
to
#include <Recorder.h> //SAJalali

2. then there is the following problem when compiling commands.cpp
```
commands.cpp(1521): error: class "Tcl_Interp" has no member "result"
  	sprintf(interp->result, "%35.8f ", res);
  	                ^

compilation aborted for commands.cpp (code 2)
make[1]: *** [../../Makefile.def:252: commands.o] Error 2
make[1]: Leaving directory '/mnt/dane/git/OpenSees/SRC/tcl'
make: *** [Makefile:50: all] Error 2
```

mhscott:
I got the same errors, @sewkokot .  Fixed them in #102 .

@sa-jalali For future reference, please use Tcl_SetResult instead of writing directly to interp->result, which has been deprecated.  You will see in tcl/command.cpp that all of these were changed to Tcl_SetResult.

sa-jalali:
Alright @mhscott . Thanks for your notice.
Thank you @sewkokot

sewkokot:
I confirm that the errors are fixed and Opensees can be built sucessfully now.
Thank you @sa-jalali and @mhscott

PR #105, opened 2019-05-21
new element recorder option: InternalForce

sa-jalali:
the "InternalForce" recorder is required to automate computation of internal forces at given locations inside the beam. This computation is done following the mechanism employed by forceBeamColumn2d element. That is, elementalLoad objects are stored at the class in addition of being used for updating the reaction and fixed-end forces. The stored loading data are then used in newly added "computeSectionForces" function to compute internal force at the intended location. This function is itself invoked by "getResponse" method of the element.

Thanks and Regards
SAJalali

### DIFF
--- a/SRC/element/elasticBeamColumn/ElasticBeam2d.cpp
+++ b/SRC/element/elasticBeamColumn/ElasticBeam2d.cpp
@@ -51,10 +51,6 @@
 #include <elementAPI.h>
 #include <string>
 #include <ElementIter.h>
-//added by SAJalali-done for InternalForce Recorder
-#define SECTION_RESPONSE_MZ		1
-#define SECTION_RESPONSE_P		2
-#define SECTION_RESPONSE_VY		3
 
 Matrix ElasticBeam2d::K(6,6);
 Vector ElasticBeam2d::P(6);
--- a/SRC/element/elasticBeamColumn/ElasticBeam2d.cpp
+++ b/SRC/element/elasticBeamColumn/ElasticBeam2d.cpp
@@ -51,6 +51,10 @@
 #include <elementAPI.h>
 #include <string>
 #include <ElementIter.h>
+//added by SAJalali-done for InternalForce Recorder
+#define SECTION_RESPONSE_MZ		1
+#define SECTION_RESPONSE_P		2
+#define SECTION_RESPONSE_VY		3
 
 Matrix ElasticBeam2d::K(6,6);
 Vector ElasticBeam2d::P(6);
@@ -224,6 +228,8 @@ ElasticBeam2d::ElasticBeam2d()
   :Element(0,ELE_TAG_ElasticBeam2d), 
   A(0.0), E(0.0), I(0.0), alpha(0.0), d(0.0), rho(0.0), cMass(0),
   Q(6), q(3), connectedExternalNodes(2), theCoordTransf(0)
+  //added by SAJalali:
+  , numEleLoads(0), eleLoads(0), eleLoadFactors(0)
 {
   // does nothing
   q0[0] = 0.0;
@@ -245,6 +251,8 @@ ElasticBeam2d::ElasticBeam2d(int tag, double a, double e, double i,
   :Element(tag,ELE_TAG_ElasticBeam2d), 
   A(a), E(e), I(i), alpha(Alpha), d(depth), rho(r), cMass(cm),
   Q(6), q(3), connectedExternalNodes(2), theCoordTransf(0)
+  //added by SAJalali:
+  , numEleLoads(0), eleLoads(0), eleLoadFactors(0)
 {
   connectedExternalNodes(0) = Nd1;
   connectedExternalNodes(1) = Nd2;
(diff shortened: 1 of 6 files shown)


sa-jalali:
@fmckenna 
@mhscott 
Please consider accepting this request. People are waiting for the new recorders to become available in OpenSees/master.

Thanks

mhscott:
This can be done using a forceBeamColumn or dispBeamColumn element with elastic sections.  I think there is unnecessary code duplication here just to get internal forces from a recorder for the elastic beam elements.

sa-jalali:
> This can be done using a forceBeamColumn or dispBeamColumn element with elastic sections. I think there is unnecessary code duplication here just to get internal forces from a recorder for the elastic beam elements.

The differences are
1. Nonlinear beamColumn's are not computationally optimum for elastic analysis
2. Nonlinear beamColumn's provide internal forces ONLY at their integration points. We desire to get forces at an arbitrary point.

Esanii:
This customization is necessary for our research. Please add it ...

AliShaani:
> 
> 
> This can be done using a forceBeamColumn or dispBeamColumn element with elastic sections. I think there is unnecessary code duplication here just to get internal forces from a recorder for the elastic beam elements.

Hello, dear Mr.mhscott
It's an honor to talk to you, on the other hand, With all respect to you, I must say this code will be help other users significantly. I strongly propose you as a researcher to check it.

SadjadGh:
Getting internal forces at "an arbitrary point" is necessary.

impce:
@fmckenna
@mhscott

The above customization is required for the research we follow. Therefore, we would be grateful if you could speed it up.

IM

nmirzai:
@fmckenna
@mhscott

I think it as an interesting option. By adding this command many researchers around the world can benefit from getting forces at an arbitrary point. We are looking forward to adding this command! Thank you.

sa-jalali:
@Esanii 
@AliShaani 
@SadjadGh 
@nmirzai 
@impce 

Thank you all for your support of this project and for your interest in the role and responsibility of our research community in international scenes.

Best Regards,
SA Jalali

sa-jalali:
Dear professor Scott (@mhscott),
We are the members of a rather large community of Iranian researchers who work together to develop their tools and utilities while working on various areas of research. Our development in OpenSees is propagated back to the international community using the github and the @civil-soft-science account. We are doing this to play our role and provide open access to our customization as others did and let us benefit from their works. What we have provided here has been employed in successful research programs and is indeed useful for other researchers working in similar areas.
While we are sure you are aware of this usefulness and while many other trivial customizations have been welcomed in OpenSees, we would like to ask why are you hesitating in accepting our contribution? Is this hesitation in favor of the OpenSees community who will benefit from these capabilities? We see the answer to be "No" and we can also see the pressures on you by the one-sided restrictive and non-constructive policies of your government against the Iranian nation. Though, you and we are researchers and researchers are in sake of truth and should avoid biased overlook of existing facts.
The existing fact is that Iranians are extremely successful worldwide in their fields of expertise and despite the propaganda of US government against them, they are extremely peaceful and other-regarding. But, they are also independent and not yielding in front of any unfair hegemony and this is what the US government do not like about us.
Another fact is that, like our countrymates who worked on peaceful nuclear sciences, our research team is also doing well and has an outstanding contribution in international publications. For sure, it is not hard for us to propagate our customized version of OpenSees independently as a free production of our team. Even, we see it not difficult to establish our version of Open-Source Finite Element software and become a serious competitor of the OpenSees team with all its weaknesses. However, our priority is collaboration and we will work with others as long as a respectful reciprocal relationship there exists. 
One more fact is that, Iranians have a strong unified community of scientists and experts that work together wherever they live. Whether inside or outside the borders of our country, we are a unified nation; not only a nation, but also a network of scientific and technological capabilities that can hardly be hindered, it at all possible.  
We hope you act more scientifically than your government! by considering these facts and by avoiding biased behaviorism that let all sides gain benefit from OpenSees.

Regards,
Seyed Alireza Jalali | Manager
CivilSoftScience

sa-jalali:
Thank you so much for your response. Meanwhile, I am glad you have checked the code and appreciate spending your time.
1. I do not remember now why I re-defined those constants. Perhaps I didn't mention that the "SectionForceDeformation.h" is included in the file and all definitions made there are available here too. Now I pushed a new commit that removes those definitions. Thanks.
2. By using the same naming convention used in "SectionForceDeformation.h", I wanted to follow the order of forces used in section output. I wanted to help the user obtain a consistent understanding of what is done in different parts of the program.

sa-jalali:
Ok. I will remove that usage.

But to clarify my intent, SECTION_RESPONSE codes determine the order in which SECTION  force components are stored in a vector and also printed in output. I wanted to adhere to that order in my code. That way, a later change in SECTION response order will also affect the order of output forces in my code.
Another thing, if you are concerned with unnecessary inclusion of .h files, please consider that the following line was originally in the file and is not part of my commit:
#include sectionForceDeformation.h

Thank you

sa-jalali:
The modifications need to be updated and the pull request may be sent again later. Thanks anyway.

PR #109, opened 2019-06-03
Update Makefiles for OSG

jasoncpatton:
Also potentially works with other HTCondor pools running Enterprise Linux (i.e RHEL, CentOS, Scientific Linux, etc.) versions 6 and/or 7.

Changes:
* Update libgfortran location 
* Update for SuperLU5.1.1
* Add CSparse

When compiling and running within an HTCondor job, users should first modify their environment to point `$HOME` to `$_CONDOR_SCRATCH_DIR` and have TCL 8.5 (with static libs) available under `$_CONDOR_SCRATCH_DIR/tcl`.

~~Note that pre-compiled `OpenSees` binaries are not easily sent along with HTCondor jobs because the search paths for `init.tcl` are hardcoded in the binary. In OSG and other HTCondor pools, users may not have access to a home directory (e.g. `/home/jasoncpatton`) and the name of the scratch directory (e.g. `/var/lib/condor/execute/slot1/dir_30502`) changes from job to job. In the case that a HTCondor pool does have a shared filesystem available, this problem can be mitigated by installing TCL 8.5 in a shared location and modifying `Makefile.def` to point to that location.~~ Jobs should set `TCL_LIBRARY`, see comment below.

### DIFF
--- a/MAKES/Makefile.def.OSG-EL6
+++ b/MAKES/Makefile.def.OSG-EL6
@@ -1,0 +1,474 @@
+############################################################################
+#
+#  Program:  OpenSees
+#
+#  Purpose:  A Top-level Makefile to create the libraries needed
+#	     to use the OpenSees framework. 
+#            and below.
+#
+#  Written: fmk 
+#  Created: 10/99
+#
+#  Send bug reports, comments or suggestions to fmckenna@ce.berkeley.edu
+#
+############################################################################
+
+
+# %---------------------------------%
+# |  SECTION 1: PROGRAM             |
+# %---------------------------------%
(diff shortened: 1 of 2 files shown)


jasoncpatton:
Regarding Tcl, a user can set an environment variable inside their job... assuming Tcl 8.5 is available under `$_CONDOR_SCRATCH_DIR/tcl`:

`export TCL_LIBRARY=$_CONDOR_SCRATCH_DIR/tcl/lib/tcl8.5`

This will allow OpenSees to run without the "cannot find init.tcl" error.

PR #115, opened 2019-07-15
Change Python output strategy

otaithleigh:
Instead of creating a stepping stone error (`opensees.msg`), we can pass things directly to Python's stderr using `PySys_FormatStderr`. This output can be caught and redirected by things expecting it (e.g. Python's `contextlib.redirect_stderr`). Additionally, the last message printed is saved, so when a real error occurs, it has that information along with it.

Finally, I've renamed `opensees.error` to `opensees.OpenSeesError` and made it available from the actual module. This allows it to be caught by the calling Python script.

### Example: uncaught error
``` python
>>> import opensees as ops
>>> ops.model('basic', '-ndm', 2)
>>> ops.node(0, 0.0)
insufficient number of arguments
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
opensees.OpenSeesError: insufficient number of arguments
```

### Example: catching the error
``` python
>>> try:
...     ops.node(0, 0.0)
... except ops.OpenSeesError:
...     print('an error occurred!')
... 
insufficient number of arguments
an error occurred!
```

### Example: capturing output with `contextlib`
``` python
>>> import contextlib
>>> import io
>>> f = io.StringIO()
>>> with contextlib.redirect_stderr(f):
...     ops.node(0, 0.0)
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
opensees.OpenSeesError: insufficient number of arguments
>>> f.getvalue()
'insufficient number of arguments\n'
```

### DIFF
--- a/SRC/interpreter/PythonModule.cpp
+++ b/SRC/interpreter/PythonModule.cpp
@@ -285,14 +285,15 @@ initopensees(void)
         INITERROR;
     struct module_state *st = GETSTATE(module);
 
-    st->error = PyErr_NewException("opensees.error", NULL, NULL);
-    PyObject* ops_msg = PyErr_NewException("opensees.msg", NULL, NULL);
+    st->error = PyErr_NewExceptionWithDoc("opensees.OpenSeesError", "Internal OpenSees errors.", NULL, NULL);
     if (st->error == NULL) {
         Py_DECREF(module);
         INITERROR;
     }
+    Py_INCREF(st->error);
+    PyModule_AddObject(module, "OpenSeesError", st->error);
 
-    sserr.setError(st->error,ops_msg);
+    sserr.setError(st->error);
 
     Py_AtExit(cleanupFunc);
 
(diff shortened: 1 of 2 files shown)


zhuminjie:
Hi,

The reason for opensees.msg is that many OpenSees codes use opserr to print information, which are not necessary "errors", and no exception should be raised. That's why opensees.msg only prints the message but not stop a function. Also due to the current error catching mechanism in OpenSees, errors often propagate from several calls to opserr. If only last error message is returned, the original error location and message are lost. Therefore, opensees.msg wants to store all information from all opserr called.

otaithleigh:
Hi Minjie,

All the messages are sent to sys.stderr. The user can inspect the output as they would before. The final message is used for the text of the error, but it too shows up in the stderr stream.

zhuminjie:
I got it for that part. But it still raises an error, right?

otaithleigh:
Yes, but only in the same situation as before.

otaithleigh:
Nodes not getting the right arguments raises an error; verbose test printing doesn't.

PR #118, opened 2019-07-30
adding some IMK, meges with master, and adding Stampede2 Makefile.def

fmckenna:


kkolozvari:
I have noticed that code for these IMK materials are very different from the one that were available previously. Just wondering what was the reason for change? I am trying to use these models, but still having a lot of issues and observing weird behavior. Please see below. 

https://github.com/OpenSees/OpenSees/issues/537#issuecomment-881044243 

Thank you

PR #145, opened 2019-10-07
Fix UVC class names and add UVCuniaxial to the TclModelBuilder

ahartloper:
c92fb2a - updated old names for the materials in the class tags
801c764 - add UVCuniaxial to the TclModelBuilderUniaxial


ahartloper:
@fmckenna  Related comment: should I also add the UVC materials to the VS64 solution?

PR #150, opened 2019-10-07
set OpenSees.sln back to VS2019

zhuminjie:


fmckenna:
why do we need 19 .. best to keep as old a .sln as possible avail to developers

zhuminjie:
I didn't set to 19. It was updated in previous commits to change to 19. I didn't know and changed to old version. I thought I should set it back to 19 again.

PR #152, opened 2019-10-09
Changing std::vector initialization

mhscott:
Changing std::vector initialization for strict ansi compiler

### DIFF
--- a/SRC/material/nD/UVCmultiaxial.cpp
+++ b/SRC/material/nD/UVCmultiaxial.cpp
@@ -22,8 +22,8 @@ static int numUVCmultiaxial = 0;
 // OPS_GetString() function: causes crash with .dll
 void* OPS_UVCmultiaxial(void) {
   if (numUVCmultiaxial == 0) {
-    std::cout << "Using the UVCmultiaxial material, see "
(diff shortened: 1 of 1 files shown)


ahartloper:
@mhscott Hi Micheal, for my own information, can you let me know more specifically what the issue was? To my knowledge, the initialization should be according to C++ standard. After I can check the other code I wrote. Thanks, Alex.

mhscott:
@ahartloper The code did not compile on Mac OS. There may also be some other small changes related to using opserr instead of cerr.  opserr should be used so that output gets redirected to the right place, e.g., in a Jupyter Notebook, cerr will go nowhere but opserr will be seen in the Notebook.

ahartloper:
OK, thank you for the reply. I'll look to compile OpenSees on my Mac as well to test there. The cerr info is good to know, I'll look to change all the std::cerr and std::cout to opserr.

ahartloper:
@mhscott I confirmed that this is a C++11 compatibility issue. This can be verified by passing the `-std=c++11` flag to g++ in the Makefile.def (default is C++98 as far as I can tell).

I agree it's best to leave the material with the changes you made for now, but using an updated version of C++ could be something to consider in the future.

PR #171, opened 2019-11-16
Elasticbeam2d

mhscott:


mhscott:
@sewkokot  :)  if else  instead of  if

PR #217, opened 2020-02-03
Minor Changes Made to GI Element Command Reading Methods

msalehi2004:
Hello Dr. McKenna or Prof. Scott,

Thank you very much for taking the time to review my files, making changes, and adding those to the OpenSees source codes. Per your request, I confirmed that the command reading methods work now, but made some minor changes to them. That said, can you please pull the updated GradientInelasticBeamColumn2d.cpp and GradientInelasticBeamColumn3d.cpp in?

Thanks,
Mohammad


mhscott:
I made the 'numData=1' change in the OPS_ functions yesterday #214 .  Your branch shows conflicts with the current code because of this.  It is best practice to merge your code with the master branch before making a pull request.

msalehi2004:
Hello Prof. Scott,

Thanks for your help; I actually noticed the changes that you had made and made my edits to your updated files. I am really sorry, but I am not sure how to merge your changes to my forked repository... Is it possible that you merge my updated files now or I have to resolve that conflict?

Best regards,
Mohammad

mhscott:
Give this a read: https://gist.github.com/CristinaSolana/1885435

Your fork is 477 commits behind the master OpenSees branch, so I'm pretty sure you haven't merged with upstream :)

msalehi2004:
Thanks Prof. Scott, I appreciate your help... I managed to merge my repo with the upstream repo and then updated my GI element codes. Could you please check and see if you may pull them in now?

msalehi2004:
Thank you Prof. Scott for your suggestion, I appreciate your time and advice.

The way that the integration method is taken in the GI element is different from the way it was defined in the FB and DB beam-column elements, as the integration type is not followed by the number of IPs and the section tags. Instead, the section tags and the number of IPs are taken as mandatory inputs, while the integration type is an optional argument. This way, we allow the users to assign different sections to the end and middle parts of the element length.

In addition, we prefer the GI element formulation to accept only the integration methods of equally-spaced IPs; so with the current command, we ensure that the users use either the default method, i.e. Simpson's, or a method among Trapezoidal or Newton-Cotes.

I hope that my explanation clarifies our reasoning. However, please let me know if you still have any concerns in this regard.

Thanks again for your help with this.

mhscott:
I understand your concern, but I think it's best from a software point of view and interchangeability with other elements to keep the same command format.  You can check the integration type in the constructor and warn users that the element should only be used with equally spaced integration points.

msalehi2004:
> I understand your concern, but I think it's best from a software point of view and interchangeability with other elements to keep the same command format. You can check the integration type in the constructor and warn users that the element should only be used with equally spaced integration points.

Sure, I will make the necessary changes and will let you know once the codes are updated.

Thanks again for your time.

msalehi2004:
Hello Prof. Scott,

I made the change that you had suggested regarding the way the integration method is defined in the Python GI element command. For the Tcl command, however, I kept the original command arguments, as was the case for the forceBeamColumn element. I think this would be necessary because there seems to exist no command for defining beam integration rules with tags for the Tcl users.

By the way, can you please let me and Dr. Sideris (@psideris) know when we the new version of OpenSees with the upgrades will be available to the public on the OpenSees website? We would like the GI element formulation to become public sooner than later so that people could try it and receive feedback. In case this could happen soon, we would be happy to provide you with the formulation description and the instructions on the element command.

Thanks again for your time.

Best regards,
Mohammad

msalehi2004:
Hello Prof. Scott and Dr. McKenna,

I was wondering if you have had a chance to review my changes, yet. I would appreciate your time.

Thanks,
Mohammad

PR #235, opened 2020-02-13
Python main

zhuminjie:
The correct way to compile a standalone python distribution for OpenSeesPy.

The script for setting up the python layout and install necessary packages will be updated later.


zhuminjie:
The purpose is to create a python distribution of OpenSeesPy something like the Anaconda

PR #239, opened 2020-02-19
Add the gradientInelasticBeamColumn source files to VS project.

ahartloper:
@fmckenna commit 0b44cec303d31d64529fee591db0df66e5dfc511 added the commands to the interpreters but the source files were never added to the Visual Studio projects.


mhscott:
Thanks!

PR #242, opened 2020-02-19
Use C++11 standard in latest MacOS Makefile.def.

ahartloper:
The AMGCL libraries included by the PFEM UMFPack solver requires at least C++11 due to the use of std::tuple. I recommend to add this flag to the Makefile.def for MacOS. Also to keep in mind for future Makefiles / the other OSes.


mhscott:
Mac compilation independently verified by @kmackiegh

PR #251, opened 2020-02-25
Fixing wrong return code in FiberSection3d::setParameter

MassimoPetracca:
The return code is wrongly initialized to 0, thus If the parameter is not handled by the section, the method will reuturn 0 (which means "handled") instead of -1.
This creates bugs in the following SSI scenario:
1) you have a beam element with fiber section
2) you have solid elements whose material stage should be updated from elastic to plastic with the updateMaterialStage command.
3) Then if the beam element has a tag < than the smallest tag of the solid elements (i.e. if it has been created before the solid elements), the updateMaterialStage will stop iterating on elements when it finds the first one who handle the paramter, in this case (erroneously) the beam element, thus not reaching the solid elements.

This PR simply changes the return code from 0 to -1 as initial value.

### DIFF
--- a/SRC/material/section/FiberSection3d.cpp
+++ b/SRC/material/section/FiberSection3d.cpp
@@ -1363,7 +1363,7 @@ FiberSection3d::setParameter(const char **argv, int argc, Parameter &param)
   if (argc < 1)
     return -1;
 
-  int result = 0;
+  int result = -1;
 
   // A material parameter
   if (strstr(argv[0],"material") != 0) {
--- a/SRC/material/section/FiberSection3dThermal.cpp
+++ b/SRC/material/section/FiberSection3dThermal.cpp
@@ -1198,7 +1198,7 @@ FiberSection3dThermal::setParameter(const char **argv, int argc, Parameter &para
     return -1;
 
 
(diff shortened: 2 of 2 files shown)


mhscott:
Does FiberSection2d do the correct thing?

Get Outlook for Android

MassimoPetracca:
FiberSection2d, FiberSection2dThermal, FiberSectionGJ, FiberSectionGJThermal, NDFiberSection2d, NDFiberSection3d and NDFiberSection2dWarping **they all do the right thing** (initialize int result = -1).

Only **FiberSection3d** and (now that I gave a second check) also the **FiberSection3dThermal** do the wrong thing.  I accidentally found this bug doing a SSI model in OpenSeesMP where only in 1 process the subdomain remained elastic even though I issued the updateMaterialStage command. That was because in that domain the beam element was added before the solid ones.

I'm going to add another commint for the FiberSection3dThermal as well

mhscott:
Thank you, @MassimoPetracca!

MassimoPetracca:
@mhscott @fmckenna , Thank you for taking the time to review this PR.

PR #258, opened 2020-03-08
fix Makefiles after moving Bilin from snap up to uniaxial

sewkokot:
This PR fixes Makefile files after moving Bilin from snap up to uniaxial.


mhscott:
Thank you, @sewkokot!

PR #262, opened 2020-03-11
Fixing bug: Null TCL Model Builder after some commands

MassimoPetracca:
Dear @fmckenna @mhscott ,  many commands (timeSeries, system, integrator, ...) call **OPS_ResetInput** passing a **NULL** tcl model builder, just because they don't have access to the global instance of the model builder. 

However this sets the pointer in elementAPI_TCL.cpp to NULL, causing a crash if another command issued later on will try to call something like OPS_GetNDM(). 

For example I experienced a crash while creating an Elastic section after a timeSeries command.

This PR simply replaces those calls to **OPS_ResetInput**, with calls to **OPS_ResetInputNoBuilder** which should be used when one does not need to change the builder (correct me if I'm wrong)

### DIFF
--- a/SRC/domain/pattern/TclSeriesCommand.cpp
+++ b/SRC/domain/pattern/TclSeriesCommand.cpp
@@ -81,13 +81,12 @@ extern void *OPS_PeerNGAMotion(void);
 
 #include <TclModelBuilder.h>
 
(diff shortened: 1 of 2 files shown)


aliirmak:
timeSeries Path is causing crash. It may be related to this issue.

PR #266, opened 2020-03-12
Fix wrong OPS calls in recent commits

zhuminjie:


mhscott:
Thank you, @zhuminjie, for cleaning up the mess I made!

PR #280, opened 2020-03-23
Fix the NULL dereference in Joint2D and Joint3D

ahartloper:
Fixes #197 

I also auto-formatted the various Joint2D/3D files for consistency.

PR #281, opened 2020-03-24
Uncomment the parsers for Joint2D/3D and fix the constructors.

ahartloper:
Fixes the issues that I made with PR #280.

### DIFF
--- a/SRC/element/joint/Joint2D.cpp
+++ b/SRC/element/joint/Joint2D.cpp
@@ -238,9 +238,10 @@ void* OPS_Joint2D()
       }
     }
 
+    UniaxialMaterial* springModels[5] = { MatI, MatJ, MatK, MatL, PanelMaterial };
     theJoint2D = new Joint2D(Joint2DId,
       iNode, jNode, kNode, lNode, CenterNodeTag,
-      MatI, MatJ, MatK, MatL, PanelMaterial,
+      springModels,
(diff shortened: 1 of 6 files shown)


ahartloper:
> Continuing the discussion from #280, why use a std::vector for DamageModels and not UniaxialMaterials? 

This was just part of my problem solving process: testing and fixing the bug, then thinking about how it could be done better the next time.

> Why not keep it simple (and presumably more portable) by using pointer arrays for both sets of objects instead of std::vector for one and not the other?

Hmm, I never thought about using arrays since to me the vector is generally simpler because of iterators (plus their other features). In my opinion they are both equally portable since it's part of the standard library.

mhscott:
It looks like it's a fixed number of DamageModels, so iterators don't buy you much.  Plus, some folks still have issues with the new standard library.  I think it's always best to not use standard libraries unless absolutely necessary.  That does not appear to be the case here.

ahartloper:
> It looks like it's a fixed number of DamageModels, so iterators don't buy you much. Plus, some folks still have issues with the new standard library. I think it's always best to not use standard libraries unless absolutely necessary. That does not appear to be the case here.

Yes true. Fair enough, I can make the change no problem.

ahartloper:
@mhscott See the latest commit for my proposed changes and let me know what you think :) .

PR #287, opened 2020-03-31
Partition in OpenSeesPy

zhuminjie:
Add a partition command in OpenSeesPy, which simulates the OpenSeesSp behavior. But user needs explicitly to call the partition command to partition the domain to subdomain on different processors after the model is built. Then the rest of setup is like OpenSeesMp, such parallel numberer and solver.


mhscott:
@zhuminjie, I get this after merge:

/home/mhscott/OpenSees/SRC/system_of_eqn/linearSOE/mumps/MumpsSolver.h:49:10: fatal error: dmumps_c.h: No such file or directory
 #include <dmumps_c.h>
          ^~~~~~~~~~~~
compilation terminated.

PR #294, opened 2020-04-05
Adding cracking output for PlaneStressUserMaterial

vpapanik:
I have added cracking output for PlaneStressUserMaterial as follows

recorder Element -file **_filename_** -time -ele **_shell element number_** material **_gauss point_** fiber **_LayeredShell concrete layer_** cracking

e.g. recorder Element -file cracks.out -time -ele 1 material 3 fiber 5 cracking

returns cracking for element one, at gauss point 3 for the 5th layer (must be concrete).

**_filename_** contains the following information for each load step

**time - crack flag 0/1 in direction 1 - crack flag 0/1 in direction 2 - crack angle (degrees)**

Angle is 0 when there is no cracking.

This will be soon incorporated into GiD+OpenSees interface for visualizing LayeredShell section cracking.


vpapanik:
will further add concrete crushing and rebar yielding damage output.

mhscott:
Fix in NDMaterial already done in #296. 

But, it's odd #296 and your PR show different line numbers.  Are you merged with upstream?

vpapanik:
I am sorry, I reported line 295 of my modified code (this pull request) instead of line 288.

vpapanik:
I am also suggesting that when cracking is not available for NDMaterial, an zero vector with three components should be returned instead of _errVector_

```
const Vector&
NDMaterial::getCracking()
{
    opserr << "NDMaterial::getCracking -- no cracking in material\n";
    static Vector ret = Vector(3);
    return ret;
}
```

mhscott:
It would be better to use the setResponse and getResponse methods for this functionality instead of polluting the NDMaterial interface with a method that most implementations will not use.

vpapanik:
I agree with you that it's not the cleanest solution from a programming point of view, however I did it this way because the LayeredShell section (for shell elements) works with many NDMaterials, so it should return the same cracking information for all layers (fibers), regardless if a layer material doesn't support cracking (it will return zero = no cracking).

This would be convenient for postprocessing utilities to display global cracking information for the whole LayeredShell section, without having to look which layers actually support cracking or not, i.e. no error will be returned - just 0 = no cracking.

Composite sections (R/C or encased steel) are examples of the above, where some materials such as concrete can crack and others such as steel cannot. On the latter, no cracking (0) will be returned which is true.

Thanks a lot for your consideration !

mhscott:
You can have the same functionality through setResponse and getResponse.  Does cracking make sense for soil models?

vpapanik:
Of course not, but they are not used in LayeredShell sections. The problem is to extract cracking information for all LayeredShell fibers, regardless if some materials cannot crack inside them. Otherwise, set/getResponse for cracking should be written for all of them, and return (0,0,0) if no cracking occurs in those materials. In that way, visual postprocessing of the whole section will be far easier.

mhscott:
You can make set/getResponse have default behavior in NDMaterial::set/getResponse for cracking.  All NDMaterials should be calling NDMaterial::setResponse if they run out of options (if not implemented, it's very easy to change and will solve a lot of future use cases like this), then the base class getResponse takes care of the rest.

I know there have been a few instances of interface pollution in the past, but my opinion is to not have more of it moving forward.

vpapanik:
I see what you mean, I will try to do it that way. I thought that pollution was the addition of the cracking request in the base NDMaterial, but this is not the case after all. I just have to add a default output there, which is (0,0,0) and write set/getResponse for the PlaneStressUserMaterial. Hope I understood your point and thank you very much for your support.

vpapanik:
Will resubmit

PR #309, opened 2020-04-14
Add new NDMaterial to OpenSeesPy

zhuminjie:


SRC/interpreter/OpenSeesNDMaterialCommands.cpp
```diff
@@ -156,6 +157,9 @@ namespace {
 	nDMaterialsMap.insert(std::make_pair("PM4Sand", &OPS_PM4SandMaterial));
 	nDMaterialsMap.insert(std::make_pair("UVCplanestress", &OPS_UVCplanestress));
 	nDMaterialsMap.insert(std::make_pair("UVCmultiaxial", &OPS_UVCmultiaxial));
+	nDMaterialsMap.insert(std::make_pair("UVCmultiaxial", &OPS_UVCmultiaxial));
```
ahartloper, line 160: I think line 160 is a duplicate of 159 in OpenSeesNDMaterialCommands.

zhuminjie: Thanks for catching that!

PR #314, opened 2020-04-15
Making eta/dt calculations conditional

mhscott:


vpapanik:
Excellent, thanks a lot !

PR #319, opened 2020-04-17
fixing compilation in visual studio win64

MassimoPetracca:
Dear @fmckenna @mhscott , due to some changes done in the last days, the compilation in visual studio win64 does not work. This is a list of changes for this PR:

1. remove references to the **telm** directory in reliability
2. add **PFEMContact2D** element in visual studio project files. (now PFEMContact2D.h is only included in BackgroundMesh.cpp, thus generating an unresolved symbol in linking...)
3. remove **_H5DRM** flag from **domain project** + remove **libhdf5.lib and libhdf5_cpp.lib** from the linker in **OpenSees project**

**Regarding the third point**:
I removed the use of the H5DRM pattern (by disabling the _H5DRM flag) just because it is enabled only in the OpenSees VisualStudio-Win64 project and in the makefile for STAMPEDE2 (nowhere else). So I thought it was left there for mistake.
In fact this forces the user that wants to compile opensees to install the HDF5 library (even though this is not specified anywhere).
Moreover, I noticed that in the link command it was expecting **libhdf5.lib and libhdf5_cpp.lib**. Those however are the libraries for **static linking. Are you sure you want to statically link to HDF5?** Even though HDF5 is BSD-based, the default precompiled versions are compiled with the SZIP compression support, that has its own license (I think is GPL, not sure if it is compatilble with OpenSees License).
In my **MPCORecorder** I also use the HDF5 library, but to overcome these problems, I just load it at runtime, if and only if the user is actually using my **mpco** recorder command. 

If you think it is necessary to have the H5DRM pattern support by default, we can think of uniforming it with my MPCORecorder, so that both of them load at runtime the HDF5 library in some static class... so that we do not create incompatibilities between different version.


vpapanik:
I also confirm #1 : telm folder has to be removed from Additional include directories and removed from project explorer.

vpapanik:
After syncing, these are some more broken references while compiling :

image

MassimoPetracca:
> After syncing, these are some more broken references while compiling :
> 
> image

Probably you did not update your repo correctly. Those files should be there. Also you should not have a problem of "not finding hdf5.h" since this PR removed the flag _H5DRM from preprocessor deinfitions

vpapanik:
Thanks, I checked again and I am even with current repo. Just changed the tcl paths in my local fork, nothing else and it doesn't compile. Will check again though.

zhuminjie:
I have the same error. I will fix it.

vpapanik:
@zhuminjie it's really strange. The files are there, includes are there, and files are not found... 
VS2019 16.5.4

mhscott:
The path may not be in the include directories of the project settings???

vpapanik:
Actually there is a different number of errors across the four configurations:

Release
image

Debug
image

Release.DLL
image

Debug.DLL
image

For instance, in 'Release', file **hdf5.h** is missing. However it is unconditionally included in **H5DRM.h.** (102) and **H5DRM.cpp.** (31). The _H5DRM directive is off, but the compiler still tries to find the hdf5.h file. Can't figure it out.

image

Moreover, in 'Debug', these include paths are missing from 'system' project:
..\..\..\src\element\PFEMElement
..\..\..\src\recorder

and from 'analysis':
..\..\..\other\csparse

MassimoPetracca:
I confirm that I was able to build the relase because I had HDF 1.10.5 installed on my machine. 
Infact the Win64/Release setting is the only one that includes the "c:\Program Files\HDF_Group\HDF5\1.10.5\include;" directory.
Even if I removed the _H5DRM flag, VS still compiles the H5DRM file because it is there in the project. Even if it will not be used in the tcl project, thus without generating LINK errors.

So I think the solution for the release build is to remove the H5DRM file. I will commit it, but before I have to fix also the debug build.

vpapanik:
This error applies to all four configurations, just checked it.
No other errors are present when the above missing paths are included (just updated them).

vpapanik:
@MassimoPetracca fixed everything, I can commit if you prefer.

MassimoPetracca:
@vpapanik ok, if you want go ahead!
The fixes I proposed above where fine for release, but I had to fix some other missing includes for PFEMElement, csparse and recorder directories. Additionally I had to fix the used platform toolset, since it was v141 almost everywhere except few project that where using the v142. Before I did not noticed this because I was working on a PC with both toolsets installed. Now I'm testing this on a new PC with a clean installation of just VS 2017 (v141 only) and found this problem.
I just fixed everything on this secondary PC, but I cannot do the pull request now, since I'm busy with compilation of OpenSees MP for windows with Cygwin now... so go ahaed

vpapanik:
@MassimoPetracca my only concern is that I have updated everything to latest VS2019. I left a draft request until I figure out if it is allowed.

MassimoPetracca:
@vpapanik I agree to update everything to VS2019, also because it was so before, so the change to 2017 in some project might have been a mistake.
However either of them is fine since v140 v141 and v142 are binary compatible, however it is better of course to make them consistent! let's wait

PR #320, opened 2020-04-17
Damage output for PlaneStressUserMaterial + J2PlaneStress fix + Resolved compilation issues

vpapanik:
- Conditional compilation of H5DRM.h and H5DRM.cpp
- Updated to latest VS2019 v16.5.4
- Fixed compilation issues in all configurations

@mhscott please comment on the necessity to update everything to the latest VS2019 version. If this is not possible please ignore this commit. However, there were quite a few compilation issues that have been fixed.


ahartloper:
Maybe we should call the `OpenSeesPy38` project as just `OpenSeesPy3`? I can easily imagine that in a few months this will be 39, etc..

vpapanik:
Also fixed the tolerance value in J2PlaceStress from absolute (1e-12) to relative.

fmckenna:
just looked at what you all wanted to check in .. with regards the HDF5DRM .. those changes should not have been done in the H5DRM class. ... the error somebody made was including  those files in the windows project solution .. my projects, which I do not check in, have them .. so when i accept this request i am going to turn around and undo those changes.

vpapanik:
@fmckenna that is even better, in order to avoid compilation issues. After hopefully merging this PR you can remove those files from the VS solution.

vpapanik:
Prof. @mhscott are there any issues with the present PR ?
Apart from the first commit, I think that the next two have no mistakes and work fine.

vpapanik:
Should I remove this PR and try again? much have changed since then and don't know what to do...

PR #321, opened 2020-04-18
Make MUMPS defaults to unsymmetric matrix type + make matrix type user defined

MassimoPetracca:
Dear @fmckenna @mhscott, I recently found out that the same analysis had **different convergence rates** in sequential and parallel versions of OpenSees (sequential using SuperLU and parallel using MUMPS).

Then I realized that the **Mumps(Parallel)SOE defaults to "general symmetric" matrix** type. While I was using a plastic damage model whose tangent matrix is unsymmetric due to the spectral decomposition of the stress tensor.

I think the **MUMPS should use Unsymmetric matrix type by default**, since it is the most generic choise (we may have unsym matrices in many cases, unilateral plastic-damage models, non-associated plasticity, beams with corotational kinematics... ).

In this **PR** I made the Mumps(Parallel)SOE set up their **matType fields to 0 (unsym)** in their constructors as default values. Then I **added the -matrixType** option in the TCL/Py commands.

I also **fixed a  bug** in the Py interpreter, in case MUMPS is used in sequential version, since there the builder command was returning NULL and not the constructed SOE.

### DIFF
--- a/SRC/interpreter/OpenSeesCommands.cpp
+++ b/SRC/interpreter/OpenSeesCommands.cpp
@@ -3156,6 +3156,7 @@ void* OPS_ParallelDisplacementControl() {
 void* OPS_MumpsSolver() {
     int icntl14 = 20;
     int icntl7 = 7;
+    int matType = 0; // 0: unsymmetric, 1: symmetric positive definite, 2: symmetric general
     while (OPS_GetNumRemainingInputArgs() > 2) {
         const char* opt = OPS_GetString();
         int num = 1;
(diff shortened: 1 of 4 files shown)


fmckenna:
in future strive to leave the defaults as is and add options to do something different .. while I would agree that unsymmetric might make more sense, it will confuse people who might start getting different answers and will annoy those whose analysis start taking 50 to 100% longer given that most materials in OpenSees actually symmetrise the tangent

mhscott:
So, did the default change or not?  I agree with @fmckenna that the default should not change

MassimoPetracca:
@fmckenna and @mhscott, I see your point, you are perfectly right that default should not change.
I can bring the default back to 2 (generic symmetric) in the SOE constructors. I already added an option to specify the matrixType in the builder command (also there I will bring the default back to 2 in case the user do not specify it).
Should I create another PR for this? Or can I just push another commit here?

PR #322, opened 2020-04-18
Improve wrapper response

MassimoPetracca:
@fmckenna @mhscott, the NDMaterial wrappers do not implement the setResponse method, thus the call falls back to the NDMaterial base class implementation, where only strain stress and tangent (and few things for thermal analysis) are supported.

In this way however if we use an adapter for a material that has additional outptus, they will not be vivisble.
In this PR we use the base class implementation for strain/stress/tangent so that we retain the strain-size of the adapter, while we forward the call to the adaptee for custom results.


vpapanik:
@MassimoPetracca looking forward for this PR to be merged in order to set up the cracking response of PSUMAT material properly !

MassimoPetracca:
> @MassimoPetracca looking forward for this PR to be merged in order to set up the cracking response of PSUMAT material properly !

@vpapanik I had a look at your request for cracking response but I cannot find it anymore. How should your "cracking" output look like? I immagine it is a vector parallel to the crack direction. Is it going to be just a Unit-vector if crack is open, and 0-vector if not? or are you going to link somehow the vector magnitude to the crack opening?

vpapanik:
@MassimoPetracca I withdrew the previous PR because it was not done properly, so I have to do it using set/getResponse.

In PlaneStressUserMaterial (PSUMAT), I wrote a method to get cracking output. It is a simple vector with three elements, namely : Crack flag (0/1) in direction 1, crack flag in direction 2 and crack angle. This is sufficient since a postprocessor can take care of the graphical part to show the crack vector. I wrote a getCracking method as follows:

```
const Vector& PlaneStressUserMaterial::getCracking()
{
    static Vector vec = Vector(3);

    vec(0) = statevdata[27];                          // crack 0/1 in direction 1

    if ((vec(0) != 0) && (vec(0) != 1)) vec(0) = 0;   // clean unwanted values

    vec(1) = statevdata[38];                          // crack 0/1 in direction 2

    if ((vec(1) != 0) && (vec(1) != 1)) vec(1) = 0;   // clean unwanted values

    vec(2) = statevdata[15]*180/3.14159;              // crack angle (degrees)

    vec(2) = (int)(vec(2) * 100.0) / 100.0;           // round to 2 decimals    

    if (vec(2) >= 360) vec(2) -= 360;                 // fix angle for over 360 degrees

    if ((vec(0) == 0) && (vec(1) == 0)) vec(2) = 0;   // zero angle for no cracking

    return vec;
}
```

This vector should show up in output when the directive 'cracking' is present. Maybe this directive can be generalized to 'damage', to support other implementations as well.

In my previous PR, I did this using ::GetCracking in nDMaterial, instead of writing a get/setResponse method. Hence, I will now have to write a default output in get/setResponse of nDMaterial.cpp that gives a zero vector (0,0,0) for all materials, and only use the specific get/setResponse in PlaneStressUserMaterial when concrete is used in LayeredShell.

So, is it possible to implement this default output in nDMaterial using the directive 'damage' ? Please tell me if there is a way to cooperate in this. Thanks for your input.

vpapanik:
This works fine at the end of setResponse on NDMaterial.cpp

```
  else if (strcmp(argv[0], "Damage") == 0 || strcmp(argv[0], "damage") == 0) {
      static Vector vec = Vector(3);
      for (int i = 0; i < 3; i++) vec[i] = 0;
      theResponse = new MaterialResponse(this, 5, vec);
  }
```

MassimoPetracca:
@vpapanik If we want to do it as general as possible I would do this:

- I would use a more generic name like "DiscontinuityPlane" or "LocalizationPlane", since "cracking" is too specific to brittle materials.
- I would use as output a 3D Vector normal to the discontinuity plane (or a 2D vector normal to the discontinuity line in 2D...)
- In the setResponse/getResponse method of nDMaterial I would write the general implementation that is valid for every materials that provide a consistent tangent tensor. This can be achieved using the Localization Analysis of the Acoustic Tensor Q = n*C*n. Where C is the tangent matrix and n is the normal to the discontinuity. Then det(Q) = 0 determines the onset of localization for that n, i.e. loss of ellipticity of the governing equation.
- Then whenever a particular material (as in your case) can compute the discontinuity explicitly, can implement its own implementation in their setResponse/getResponse method.

The reason why I would use the normal to the crack rather then the crack direction itself (they are of course orthogonal) is because in 3D the crack does not have a direction, in fact it is a plane, hence the most compact entity that can represent it is its normal vector.

Then the Post-Processor should be able to represent a Glyph for a plane given its normal. In 2D things are easier since you just need to draw a line orthogonal to that normal. In 3D instead you would need to draw a Circlular surface orthogonal to the normal vector.

If you agree with the above solution I can start writing the implementation of the localization analysis in the base NDMaterial class (it will take a couple of days to do the necessary tests). Then in your case you just need to output the normal to the crack direction given by your PSUMAT

vpapanik:
@MassimoPetracca the problem with PSUMAT and similar plane cracking materials is that it has either zero, one or two distinct cracks in orthogonal dimensions, so it is not right to output only one plane or one line. My implementation has only to do with damage output in LayeredShell element, so it's not very generalized. So I suggest we can do both implementations since there is no conflict between them.

MassimoPetracca:
@vpapanik I see, so ok start with your implementation. I will do mine later on.

vpapanik:
thanks @MassimoPetracca, we will work things out !

vpapanik:
@MassimoPetracca everything is done from my side, you can check the updated PR.

fmckenna:
in relation to the changes to the setReponse() method ..  I really do not get the purpose of your proposed changes .. any material can override the setResponse  .. all you have added is code that checks for the base class methods and if not there invokes setResponse on the derived material  .. is the derived material provides a setResponse, the derived materials method will 
be invoked anyway and not the one you provide .. so the method provided seems quite pointless, in fact it involves (stage 2 change) some extra strcmp and a method invocation for materials that provide no setResponse()!

mhscott:
I think the point is that the wrapped class may provide a setResponse that will not get called unless the logic is changed in the wrapper class.  If the wrappers do nothing and default to the base class, then the wrapped class setResponse never gets called.

Please correct me if I'm wrong.

mhscott:
But, at any rate, this PR cannot be accepted because you have changed the third argument in the signature of setResponse.

MassimoPetracca:
> I think the point is that the wrapped class may provide a setResponse that will not get called unless the logic is changed in the wrapper class. If the wrappers do nothing and default to the base class, then the wrapped class setResponse never gets called.
> 
> Please correct me if I'm wrong.

@mhscott  **Exactly**, that is the point of this PR. 
All the wrappers (plane stress, plane strain, and plate fiber) do not forward the setResponse call to the wrapped class. Thus the wrapped class setResponse method never gets called!

This PR does exactly that. **The wrapper forwards the call to the wrapped** class.

However, I first check the response type. If it is **stress**, **strain** or **tangent**, I call the base class NDMaterial setResponse, otherwise I forward the call to the wrapped class, so that the user has the **correct strain-size of the wrapper** (and not that of the wrapped), which is **the default now**.

Example:
- Wrapped = 3D material (strain size = 6)
- Wrapper = 2D plane stress (strain size = 3)

Now:
- setResponse("strain") -> output = Vector(3) (not forwarded to the wrapped)
- setResponse("some custom result available in the wrapped") -> output = nothing (not forwarded to the wrapped)

With PR:
- setResponse("strain") -> output = Vector(3) (not forwarded to the wrapped)
- setResponse("some custom result available in the wrapped") -> output = "some custom result" (forwarded to the wrapped)

MassimoPetracca:
Dear prof. @mhscott , I received an e-mail notification with the following concern:

> But, at any rate, this PR cannot be accepted because you have changed the third argument in the signature of setResponse

Is there anything wrong with this PR? I don't think I changed the NDMaterial setResponse signature. Or did I miss something?

Unfortunately without this PR, the only results that we can see from 3D materials wrapped as PlateFiber/PlaneStress/PlainStrain are just strain/stress/tangent.

vpapanik:
I confirm the above. It is what I did as well for PlateForPlaneStress material, also adding the default damage output in NDMaterial (zero vector).

mhscott:
Yeah, sorry about that.  I misinterpreted the function signature.

PR #323, opened 2020-04-18
Adding the Plugin-Framework

MassimoPetracca:
Dear @fmckenna @mhscott, in this PR I would like to add the **Plugin Framework**.
The Plugin Framework allows writing **external (plugin) shared libraries** loaded at run-time that can contain one or multiple components (such as materials, elements, etc...).

Currently, the standard way to add a new procedure in OpenSees, requires the user to link statically to some opensees code. This can create **ABI incompatibility problems** and it is hard to maintain.

With this plugin framework the user just need to **include a single header file "PluginFrameworkAPI.h"**. And this is the only file that should be carefully maintained to avoid breaking backward/forward ABI compatibility.

I already tested it with **OpenSees/OpenSeesSP/OpenSeesMP** compiled in Windows with the Cygwin toolchain, while the plugin shared libraries are built with VisualStudio, without creating any incompatibility.

In this first PR I add the framework and support for uniaxial and NDMaterial plugins. Later on I will add support for Elements as well.

Finally I will add some **example projects** that show how to create a plugin library written either in **C, C++ or even in Python**, that can be useful for fast prototyping of new components.


fmckenna:
you will need to explain the purpose for this in greater detail .. nobody other than a certain company uses the .dll approach to distribute code .. the only other users of the shared library approach are developers testing new classes: here they build their own OpenSees exe and create shared libraries using that code so linking time is trivial .. the ABI is not an issue .. the ABI is an issue if we alled developers to put up their contributions by only proving shared libraries .. this is counterproductive to the open source nature of OpenSees .. developers who want to keep their code private at present have to at least share it with me and even then they need good reasons to keep it private otherwise i won't add it to the exe.

ps. simple API you are proposing constantly break as other developers need access to things not in the interface, e.g. material developers who want access to the domain or to the element that is using them, or elements that need access to other elements .. in 20 years of doing this i am always surprised at least once a year about what some people want to get access to given the classes they are dealing with.

mhscott:
I agree with @fmckenna.  Features specific to your purposes, features that generally break encapsulation or some other tenet that OpenSees is built on, should not become everyone's burden.  You can consider maintaining your own fork of OpenSees and adding layers, APIs, etc. that you need for whatever it is you're doing.

MassimoPetracca:
> this is counterproductive to the open source nature of OpenSees

> should not become everyone's burden

It makes perfectly sense. I understand that a software that is opensource should discourage incorporation of code that has few or nothing to do with the opensource philosophy.

> the ABI is an issue if we alled developers to put up their contributions by only proving shared libraries

Since I've seen that there is already the concept of providing components via shared libraries, I thought that it was meant to be in OpenSees. Therefore since I had truble in the past using the WrapperNDMaterial (due to differences in the source code in "DEVELOPER" and in "SRC") I thought it was a good idea to provide a lighter alternative.

So, @fmckenna  @mhscott , If you agree I will close this PR and keep this branch on my fork if it will ever get useful in the future.

PR #342, opened 2020-05-11
bugfix normalRV

talledodiego:
This PR contains a possbile bugfix in the evalaution of the PDF of normal randopm variable.
Now, compared with octave it gives the same result.


mhscott:
Thank you @talledodiego for finding this.  Did you also check the CDF and inverseCDF functions?

talledodiego:
@mhscott I check CDF and is ok. I did not check inverseCDF. I will check it today.

talledodiego:
@mhscott I confirm CDF and inverseCDF are OK. Indeed they use the normal standard distribution and transform the variable, so it works.

PR #343, opened 2020-05-17
Adding integrator to get M, C, and K

mhscott:


rorenzosan:
Kind professor Scott,
I am using Opensees, but I am not very experienced. I would like to know how to add the integrator to the Opensees path, to be able to use it.

mhscott:
It should be in the next release of OpenSees.exe.

PR #360, opened 2020-06-12
ExpressNewton algorithm

yulee:
ExpressNewton algorithm https://www.researchgate.net/publication/342098037_An_efficient_and_unconditionally_stable_numerical_algorithm_for_nonlinear_structural_dynamics


mhscott:
How is this different from using ModifiedNewton with maxIter of the convergence test set equal to the number of iterations, then telling the algorithm to proceed (I think with printFlag=5 in the convergence test or something like that)?

yulee:
Good question, Michael.

There are two differences:
1. ModifiedNewton performs factorization in every iterations even if "-initial" is activated.
2. ModifiedNewton does not scale the stiffness matrix. When this is combined with a constant number of iterations, it becomes unstable for hardening systems.

If these two items are revised in ModifiedNewton, then we get exactly the same performance. To avoid adding more parameters to ModifiedNewton that may be confusing to other users, we took the liberty and put together the new algorithm.

mhscott:
Yuli,

Thanks for the explanation.  For 1. do you mean every time step instead of every iteration?  For 2. Modified Newton is stable for hardening systems if you give it enough iterations to converge.

I'd be interested in reading more about the scaling in the paper you referenced.

Michael

yulee:
Hi Michael,

    Yes, every time step. You are right in that ModifiedNewton is stable if iterated to convergence. One of the keys for efficiency is the use of constant number of iterations which unfortunately conflicts with ModifiedNewton as is for stability reason.

Best,
Yuli

yulee:
Hi Michael,

Yes, every time step. You are right in that ModifiedNewton is stable if iterated to convergence. One of the keys for efficiency is the use of constant number of iterations which unfortunately conflicts with ModifiedNewton as is for stability reason.

Best,
Yuli

PR #365, opened 2020-06-17
Add ASD Shell Q4 Element

MassimoPetracca:
Dear @fmckenna @mhscott, with this PR I would like to add a new shell element to OpenSees ans OpenSeesPy.

It is a general purpose 4-node thick shell element with the following features:

- The membrane behavior is enhanced with the AGQ6-I formulation, that makes the element almost insensitive to geometry distortion, as opposed to standard iso-parametric elements.
- The drilling DOF is treated with the Hughes-Brezzi formulation, with special care to avoid membrane locking, using a 1 point quadrature plus stabilization.
- The plate bending part is treated using the MITC4 formulation, to avoid the well known transverse shear locking behavior of thick plate elements
- It can be used to model both flat and warped geometries
- Kinematics can be either linear or corotational. The corotational kinematics are based on the work of Felippa et al., i.e. the EICR (Element Independent Corotational formulation). Finite rotations are treated with Quaternions.

**Note:**
The element command is pretty simple:
`element ASDShellQ4 $ele_tag $n1 $n2 $n3 $n4 <-corotational>`
The last optional argument activates nonlinear kinematics.

**Question:**
@fmckenna @mhscott, How can I add the documentation for this element? Should I use the new Documentation on GitHub? We also prepared validation files for almost all NAFEMS benchmarks for shells in linear and nonlinear kinematics. We can share them too.

**Some examples:**

**Can Crush:**
_Implicit dynamic analysis of a can crush, with large displacements and contact_
Can_Crush_reduced (1)

**Bending roll-up:**
_This example assesses how very large rotation are handled. Here a 4xfull rotation(4 * 2 * &#960;) is imposed at the beam free end, and the solution converges in 20 time steps (5 newton iterations for each step), thus handling a 0.4 * &#960; incremental rotation per step_
rollup

**Mesh-bias dependency for fracture propagation**
_A simple notched plate is pulled along the in-plane vertical direction. It uses a Rankine damage model, so the expected crack should be horizontal (orthogonal to the tensile stress). With strain-localization the crack tends to propagate along the mesh direction when standard iso-parametric elements are used. The enhancement in the membrane behavior of the ASDShellQ4 improves the prediction of the crack direction_
damage


fmckenna:
do the documentation on github .. that is where the documentation pages are going to link to after hopefuilly my summer holiday (that time i will be spending finishing those pages off!)

PR #396, opened 2020-07-13
Fix #393.

### LINKED ISSUE #393: Linking errors in Visual Studio
Dear all, 
@fmckenna @mhscott , currently the visual studio projects does not compile due to some linking errors about:

- DegradingPinchingBW
- ElementRecorderRMS
- NodeRecorderRMS

Their source are missing in the VS projects, but the OPS_*** methods are called, thus the link error. Should those OPS_*** methods be removed, or should their source code be added to VS project files?

ahartloper:
- Add source files for RMS recorders
- Add source files for DegradingPinchedBW uniaxial material


aschellenberg74:
already fixed the Win64 and Win32 project files

PR #400, opened 2020-07-17
Big merge of some of my stuff

jaabell:
**This needs to be reviewed because I touch some of the base classes.** 

This is stuff I have been working on for the past year. The biggest deal is a new technique to do incremental construction analysis which is suitable for geotech applications and also fully compatible with parallel processing mode.


jaabell:
Oh wow! Just like that!! Boom!

PR #403, opened 2020-07-22
Add the SLModel

ahartloper:
- Add the SLModel by Suzuki and Lignos, see:  https://doi.org/10.1002/eqe.3324
- Add the MAT_TAGs for IMKPeakOriented and IMKPinching that were previously missing

@fmckenna How should we handle the documentation? As I understand you are moving everything from the wiki to OpenSees/OpenSeesDocumentation, so I imagine it's best just to put it there directly?


mhscott:
Yes, please add the documentation to OpenSees/OpenSeesDocumentation.  Thanks!

PR #441, opened 2020-09-16
fixing compilation in VS

MassimoPetracca:
@fmckenna @mhscott 
With the latest changes, **VS projects do not compile**, due to missing files in VS projects and a misuse of a local variable in constant expressions.

The following classes were missing in the VS project files:
- **StagedLoadControl** (integrator)
- **StagedNewmark** (integrator)
- **SixNodeTri** (quad) [I know it is a triangle, but the author put those files in the fourNodeQuad folder, so I put them in the quad subfolder of the element project file]

The following class has a compilation error due to a misuse of a class **member variable in constant expressions**. I simply changed the declaration adding an inline initialization.
- SixNodeTri

### DIFF
--- a/SRC/element/fourNodeQuad/SixNodeTri.cpp
+++ b/SRC/element/fourNodeQuad/SixNodeTri.cpp
@@ -151,9 +151,6 @@ SixNodeTri::SixNodeTri(int tag, int nd1, int nd2, int nd3, int nd4,
 	b[0] = b1;
 	b[1] = b2;
 
-	nip = 3;
-	nnodes = 6;
-
     // Allocate arrays of pointers to NDMaterials
     theMaterial = new NDMaterial *[nip];
 
--- a/SRC/element/fourNodeQuad/SixNodeTri.h
+++ b/SRC/element/fourNodeQuad/SixNodeTri.h
@@ -132,8 +132,8 @@ private:
 
   Matrix *Ki;
 
-  int nip; // number of integration/Gauss points
-  int nnodes; // number of nodes
+  static constexpr int nip = 3; // number of integration/Gauss points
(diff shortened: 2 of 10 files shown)


sewkokot:
My fault, however gnu c++ compiles successfully ;)

In SixNodeTri.cpp I hope the only place affected by this compilation error is in the following line

 https://github.com/OpenSees/OpenSees/blob/d09d8aac02a41634a8e7b59efa60fa092516e40e/SRC/element/fourNodeQuad/SixNodeTri.cpp#L1239

The array size can be hardcoded here into `We[6][3]`

In my another PR (https://github.com/OpenSees/OpenSees/pull/440) with similar situation, I have just pushed a commit to make the sizes of arrays hardcoded. 

Regarding the placement of SixNodeTri triangle element, in fact, it was easier to start with the infrastructure of quad directory than the Tri31.cpp in triangle directory. I assumed the starting point for ``SixNodeTri.cpp` which was `FourNodeQuad.cpp` is better maintained than `Tri31.cpp`, and therefore `SixNodeTri.cpp` landed in quad directory in my PR.

MassimoPetracca:
@sewkokot Yes the only affected line is that one. 
However I proposed in this PR to make your two member variables "**nnodes**" and "**nip**" **static** and **constexpr** (with an inline initialization... we can use c++11 here) also because you use them in many for-loops (loops on nodes, loops on gauss points...) and having a variable that results in a constant expression can help the compiler to optimize that code, for exaple with loop-unrolling.

sewkokot:
@MassimoPetracca thank you, I have learn something new :) So I improved PR #440 to avoid hardcoded values and reuse `nnodes` and `nip` variables.

PR #453, opened 2020-10-07
Concrete02mod material

juandiegopozo:
This new material is based on the Concrete02 material. However, it defines the pre-peak response based on the Popovics equation instead of the Hognestad parabola. Therefore, the Young's modulus is required as an input.

### DIFF
--- a/SRC/material/uniaxial/Concrete02.cpp
+++ b/SRC/material/uniaxial/Concrete02.cpp
@@ -35,6 +35,17 @@
 // by D. Sze and Filip C. Filippou in 1994
 //-----------------------------------------------------------------------
 
+//-----------------------------------------------------------------------
+//Modified on 2020 by:
+// Juan D. Pozo (jpozo@nd.edu)
+// Department of Civil and Environmental Engineering and Earth Sciences, University of Notre Dame, Notre Dame, Indiana, USA and 
+// Department of Structural and Geotechnical Engineering, Pontificia Universidad Catolica de Chile, Macul, Santiago, Chile
+// Yahya C. Kurama (ykurama@nd.edu)
+// Department of Civil and Environmental Engineering and Earth Sciences, University of Notre Dame, Notre Dame, Indiana
+// Matias A. Hube (mhube@ing.puc.cl)
+// Department of Structural and Geotechnical Engineering, Pontificia Universidad Catolica de Chile, Macul, Santiago, Chile
+//-----------------------------------------------------------------------
+
 
 #include <stdlib.h>
 #include <string.h>
--- a/SRC/material/uniaxial/Concrete02.cpp
+++ b/SRC/material/uniaxial/Concrete02.cpp
@@ -56,7 +56,7 @@ OPS_Concrete02()
   UniaxialMaterial *theMaterial = 0;
 
   int    iData[1];
-  double dData[7];
+  double dData[8];      //jdPozo
   int numData = 1;
 
   if (OPS_GetIntInput(&numData, iData) != 0) {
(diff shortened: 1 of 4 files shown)


juandiegopozo:
Hi, I would like to describe better the benefit of this new material. 

The original Concrete02 defines the pre-peak behavior with the Hognestad parabola that depends on the peak strength and the corresponding strain (eo,fc). In this behavior the Young's modulus is implicitly defined as E=2*fc/eo. In the case of confined concrete, the peak strength, fcc=K*fc, and the corresponding strain eco=eo*[1+5*(K-1)] are increased due to the lateral confinement by a confinement increase factor K. This confinement increase factor can be calculated according different authors (i.e., Mander et al., 1988, Saatcioglu and Razvi 1992). 

The problem with Concrete02 arises because the Young's modulus, E, and the strain at peak strength, eco are dependent in the Hognestad parabola and cannot be independently defined. E is required to be matched so the initial stiffness (that impacts the period of the model) is adequately estimated, while eco is required to be matched so the ductility is correctly estimated. When using Concrete02 the user has the option to either match E or eco to specific values but not both. If E is matched, eco must be defined as eco=2fcc/E giving a value that generally is less than the theoretical value of eco=ec*[1+5*(K-1)] (see red plot on attached figure). On the other hand, if eco is matched, E is implicitly defined as E=2*fcc/eco which ends in a value that is generally less than the Young's modulus defined for the unconfined concrete (E=2*fc/eo), which is unrealistic (see blue plot in attached figure).

Therefore, the proposed Concrete02mod improves the original Concrete02 material by defining the pre-peak response based on the Popovics equation so E and eco are not dependent and the user can match both at the same time (see black line in attached figure). Therefore, the period and ductility of the model can be adequately estimated.

The attached figure is just an example of the problems when using Concrete02 and the benefit of the proposed Concrete02mod. This figure was obtained with the following parameters:
fc=41.7 MPa
E=4700*sqrt(fc)=30350 MPa (according ACI 318) 
K=1.49
fcc=K*fc=62.13 MPa
eco=0.002*[1+5*(K-1)]=0.0069

Using Concrete02 to match E (red line):
eco=2*fcc/E=2*62.13/30350=0.00409 which is 40% less than the theoretical value of eco=0.0069
Using Concrete02 to match eco (blue line):
E=2*fcc/eco=2*62.13/0.0069=18008 MPa which is 40% less than the unconfined concrete Young's modulus according ACI 318.


Concrete02

juandiegopozo:
Thank you for your response professor Scott. 

The user can adjust the inputs to Concrete02 to match or the Young's modulus or the strain at peak strength (i.e., red or blue lines), but not both. This is because E is not an input in Concrete02 and it is implicitly defined as E=2*fcc/eco following the Hognestad parabola.

The modification uses Popovic's equation, thus the user is also required to define E as an input. In this case both the Young's modulus and the strain at peak strength can be matched (i.e., black line).

mhscott:
Can you define a new constructor for Concrete02?

I'm sorry for the questions, but it would be good to re-use as much code as possible.  This seems to be simply an input parameter issue, not an issue of the Concrete02 formulation.

juandiegopozo:
Yes, it will work with a new constructor.

My plan is to add E as an optional input to Concrete02 and modify also the compression envelope subroutine to use Popovic's equation if E is given as an input.

Would that be better for you professor?

juandiegopozo:
Dear professor Scott,

I have already added a new Constructor for Concrete02 and updated the pull request. The Concrete02mod files should not considered anymore since the changes were incorporated into the original Concrete02 code.

Please let me know if you have any questions.

Regards,
Juan Diego

mhscott:
Sorry for the delay, but how do the modifications you propose to the Concrete02 envelope differ from what's in Concrete04?

I guess the question is how does Concrete02mod differ from Concrete04?

juandiegopozo:
Hi professor Scott,

The most important difference is that Concrete02mod uses Popovic's equation only for the pre-peak response. The post-peak response of Concrete02mod is linear, thus allowing regularization of the ultimate strain based on an energy approach (Coleman and Spacone 2001). Regularization of concrete in compression has been used as an approach to reduce the mesh-sensitivity of RC elements showing a softening response (Pugh et al. 2015, Vasquez et al. 2016).

On the other hand, Concrete04 uses Popovics equation for the pre-peak and pot-peak responses and regularization cannot be performed.

mhscott:
Thanks for the information.  Please take a look at Concrete02IS.  I think that does exactly what you want.

juandiegopozo:
You are right. Thank you for pointing it out, I didn't see that material before.

PR #458, opened 2020-10-16
Four element  commands have been added 

jiandong-qiao:
Four element  commands have been added  in OpenSeesUniaxialMaterialCommands.cpp


fmckenna:
you need to add the commands to the upcoming documentation

also need an example in EXAMPLES/ExampleScripts that we can test

fmckenna:
you need to add the commands to the upcoming documentation

also need an example in EXAMPLES/ExampleScripts that we can test

fmckenna:
this was accepted in another pull request

PR #477, opened 2020-12-08
Element Length Computation

p-ranchal:
The length of [ElasticTimoshenkoBeam](https://opensees.berkeley.edu/wiki/index.php/Elastic_Timoshenko_Beam_Column_Element) elements is currently being computed using nodal coordinates instead of coordinate transformation, which I believe is inconsistent with the [elasticBeamColumn](https://opensees.berkeley.edu/wiki/index.php/Elastic_Beam_Column_Element) element.

[This](https://portwooddigital.com/2020/11/19/rigid-joint-offsets/) blog post reminded me that I'd come across this issue some time ago while exploring OpenSees' implementation of rigid joint offsets. However, I wanted my beams and columns to exhibit rigid behavior only in flexure and not in axial dofs. Thus, the changes in this PR only reflect the use of `Lf` in flexure stiffness terms. 

While the use of two length parameters - `L` and `Lf` - sidesteps the problem associated with applied member loads noted in the blog post linked above, it is inconsistent with the implementation of the elasticBeamColumn element (which has only one length parameter).

Also added the `.md` extension to the README file for GitHub [markdown](https://guides.github.com/features/mastering-markdown/) formatting.

Look forward to your comments on this implementation and on how to proceed.


mhscott:
Thanks for catching this.  For consistency with the other frame elements, L from the coordTransf should be used for all calculations.  Can you update to just use L instead of L and Lf?

p-ranchal:
@mhscott all done 👍

PR #485, opened 2020-12-18
Add Modal Properties and Response Spectrum Analysis

MassimoPetracca:
Dear @fmckenna, @mhscott 

With this PR I would like to add two new commands to OpenSees.

**1 - Computation of Modal Properties**
Command: **modalProperties** <-print> <-file $reportFileName> <-unorm>
This command should be called after an **eigen** command, and it will compute:

- Total and free mass of the structure
- Center of mass (to include the effect of translational masses gyrating about the center of mass)
- Modal participation factors
- Modal participation masses (for each mode, cumulative and ratios)

It takes into account the entire mass matrix of the model, thus not limited to nodal masses.
The mass matrix is stored with a sparse (triplet) storage.
The mass matrix can be either consistent or lumped. If the mass is consistent, this command additionally computes a diagonalized version of the mass matrix for the computation of the center of mass.

Additionally, the computed modal properties can be printed to the console (-print option) and/or to a file (-file $fileName options)

**2 - Response Spectrum Analysis**
Command: **responseSpectrum** $tsTag $dir <-scale $scale>
This command should be called after an **eigen**/**modalProperties** command pair, and it will compute **N** analysis steps, where N is the number of eigenvalues requested in the eigen command.
For each step, it will compute the modal displacement for the i-th eigenmode.
It does not perform any modal combination, that is left to the user since it can be easily done in post-processing.

- **$tsTag** is the time-series tag used as a response spectrum
- **$dir** (1 to NDF) is the DOF index the response spectrum should act on.


sewkokot:
Massimo, thank you for this PR.

I wanted to test this PR locally and when importing the compiled Python opensees.so module I get the following error:

`ImportError: opensees.so: undefined symbol: ops_getanalysismodel_`

MassimoPetracca:
> Is this error related to the following line present in both DomainModalProperties.cpp (line: 382) and ResponseSpectrumAnalysis.cpp (line 73)?
AnalysisModel* theAnalysisModel = *OPS_GetAnalysisModel();

You're right, that function is only available in the elementAPI_TCL. I will fix it soon.

MassimoPetracca:
@sewkokot Here it is, I created a definition of OPS_GetAnalysisModel also in SCR/interpreter/OpenSeesCommands.cpp.
I just tested it with OpenSeesPy project using Python 36 64 bit.
Let me know if I can help you with the response spectrum analysis or if you have some suggestions about it

MassimoPetracca:
I have a question for @fmckenna  and @mhscott , regarding the documentation of the new features using the new Documentation on GitHub, do I need to make a pull request for the documentation right now? or should I wait for this PR to be merged?

sewkokot:
@MassimoPetracca Thank you for the quick fix. I can confirm, no import error and the modalProperties command works!

> Let me know if I can help you with the response spectrum analysis or if you have some suggestions about it

I also tested the responseSpectrum analysis, it seems to work however would it be possible to run the responseSpectrum command in a 'for' loop over the modes?
I get the modal displacements for each mode shape in the recorder output file, but it would be good to collect the displacements or other output quantities (e.g. section forces) directly into numpy arrays in the opeenseespy script to do the  post-processing modal combination.

I don't know, maybe instead of a standalone 'responseSpectrum', it can be implemented as adding a 'responseSpectrum' integrator as Michael (@mhscott) did for the GimmeMCK command [1] and then use the 'analyze(1)' in a loop over the mode shapes?

I'm facing a similar approach problem when implementing a new functionality calculating the steady-state response due to harmonic loading (see #498). At first I thought I would follow the path of adding a standalone command as 'eigen' or 'responsSpectrum', but now I'm thinking of adding a new integrator, since it's basically the static problem with just modified stiffness matrix (K - p^2*M) u0 = F0. What do you think?

[1] https://portwooddigital.com/2020/05/17/gimme-all-your-damping-all-your-mass-and-stiffness-too/

MassimoPetracca:
> I also tested the responseSpectrum analysis, it seems to work however would it be possible to run the responseSpectrum command in a 'for' loop over the modes?

I did not think about it, because I do the post-processing (modal combination) when all modal displacements/forces/etc.. are already recorded, i.e. when the analysis is finished.

I gave a webinar about this using OpenSees and STKO:
[Youtube Link Here](https://www.youtube.com/watch?v=vUXKVm5_rn0&feature=youtu.be)
[Webinar Link Here](https://asdeasoft.net/?stkowebinars-for-opensees) (Learning Opensees: Eigenvalue and Response Spectrum Analyses)

I could implement what you said, but without making it an integrator, I can just add an optional argument to the responseSpectrum command:
responseSpectrum $tsTag $dir <-scale $scale> **<-mode $mode_id>**

where mode_id is an integer ranging from 1 to number_of_modes

So you could use it like this:
```python
for i in range(num_eigen):
   responseSpectrum(tsTag, dir, "-mode", i+1)
   # grab your output here and do post proc
```

@sewkokot what do you think about it?

sewkokot:
@MassimoPetracca, thank you for the links, I will check them out.

> I could implement what you said, but without making it an integrator, I can just add an optional argument to the responseSpectrum command:
> responseSpectrum $tsTag $dir <-scale $scale> <-mode $mode_id>

The '-mode $mode_id> option for 'responseSpectrum' would definitely solve the problem :) I'm looking forward to it, thank you.

mhscott:
Please explain why this type of analysis needs to be implemented in C++ instead of via scripts, i.e., what do these analyses do that cannot be done in a script?

You will not be able to handle all the special cases people will think of for their particular applications, so scripts are a better approach because users can make modifications to suit their own needs.

MassimoPetracca:
> Please explain why this type of analysis needs to be implemented in C++ instead of via scripts, i.e., what do these analyses do that cannot be done in a script?

Dear Prof. Scott, 
I already implemented it via TCL scripting ([link here](https://asdeasoft.net/forum/viewtopic.php?f=40&t=1513)), it was working but I faced some problems:

- From TCL I can only access nodal masses, so all the element distributed masses are neglected
- Using the printA was not a viable solution due to the requirement of using a dense solver
- Even when using the printA in cases where I'm using elements with a consistent mass matrix (mainly 3D solid elements), I eventually need to diagonalize the consistent mass matrix to compute the center of mass and the effect of translational masses gyrating about the center of mass. This is much easier to do element-wise, rather than on the whole global mass matrix.

>You will not be able to handle all the special cases people will think of for their particular applications, so scripts are a better approach because users can make modifications to suit their own needs.

I absolutely agree with you on this point, in fact, I always try to do things via scripts first. Here I could not, so I decided to put it in C++.
In order to make it as flexible as possible, I decided to split the computation of modal properties and the response spectrum analysis into distinct commands.
For the same reason, I only compute the modal displacements in the responseSpectrum, without doing any modal combination, otherwise, I would have created limitations for the users in terms of the output they want to request.

I hope this helps you understand my approach.

MassimoPetracca:
@sewkokot Here it is, now you have an extra option for the response spectrum for doing 1 mode at a time.
Command: **responseSpectrum** $tsTag $dir <-scale $scale> **<-mode $mode_id>**
where **$mode_id** is the 1-based mode index (1 to number of eigenmodes)

```python
tsTag = 2 # time series tag for the response spectrum function (path time series)
dir = 1 # excited DOF (1 to NDF)
scale = 1.0 # optional (placeholder for future use: scales the output modal displacement)
output_node = 1
#responseSpectrum $tsTag $dir -scale $scale
for i in range(7):
	print("responseSpectrum for mode {}".format(i+1))
	responseSpectrum(tsTag, dir, "-scale", scale, "-mode", i+1)
	print(nodeDisp(output_node))
```

sewkokot:
@MassimoPetracca thank you, it works well. I tested it on a 2-bay and 7-story plane frame (Sap verification example) with 7 modes  included in the response spectrum analysis (based on ElCentro response spectrum function). I've got that the SRSS modal combination results agree 100% for displacement as well as axial forces and bending moments.

It's up to @mhscott and @fmckenna to merge this PR, but I like it and highly recommend it :)

MassimoPetracca:
Dear Prof. @mhscott, thank you for accepting this PR. I'm creating a PR on the OpenSeesDocumentation right now for the **eigen** (it was missing), **modalProperties** and **responseSpectrum** commands.

mhscott:
No problem. @fmckenna wanted to make sure the command handled constraints properly before we merged the PR.

ambaker1:
Fantastic new feature. Thank you @MassimoPetracca for adding this.

MassimoPetracca:
Dear @fmckenna  @mhscott , I just create a PR for the documentation here: https://github.com/OpenSees/OpenSeesDocumentation/pull/13

PR #497, opened 2021-01-11
BoucWen recorders

salsessa:
I modified BoucWenMaterial.cpp and BoucWenMaterial.h by introducing the methods setResponse and getResponse in order to activate the recorders.
I am wondering if it is possible to merge these files in the main repository.
Thank you
Salvatore Sessa


mhscott:
The recorder options are already handled in the base UniaxialMaterial::setResponse and ::getResponse.

PR #516, opened 2021-01-28
String element recorder

mhscott:
Converting int and double arguments to string for parameters and element recorders

### DIFF
--- a/SRC/interpreter/OpenSeesOutputCommands.cpp
+++ b/SRC/interpreter/OpenSeesOutputCommands.cpp
@@ -386,10 +386,16 @@ int OPS_eleResponse()
     numdata = OPS_GetNumRemainingInputArgs();
     if (numdata > 0) {
 	const char** argv = new const char*[numdata];
+	char buffer[128];
 	for (int i=0; i<numdata; i++) {
-	    argv[i] = OPS_GetString();
+	  argv[i] = new char[128];
+	  // Turn everything in to a string for setResponse
+	  argv[i] = OPS_GetStringFromAll(buffer, 128);
 	}
 	const Vector* data = theDomain->getElementResponse(tag, argv, numdata);
+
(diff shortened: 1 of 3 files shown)


SRC/interpreter/OpenSeesOutputCommands.cpp
```diff
@@ -386,10 +386,16 @@ int OPS_eleResponse()
     numdata = OPS_GetNumRemainingInputArgs();
     if (numdata > 0) {
 	const char** argv = new const char*[numdata];
+	char buffer[128];
 	for (int i=0; i<numdata; i++) {
-	    argv[i] = OPS_GetString();
+	  argv[i] = new char[128];
```
zhuminjie, line 391: You don't have this new char. The return value from OPS_GetStringFromAll is actually a pointer to buffer passed in.

SRC/interpreter/OpenSeesParameterCommands.cpp
```diff
@@ -293,38 +293,14 @@ OPS_Parameter()
     if (numr > 0) {
 
 	char** argv = new char*[numr];
-
+	char buffer[128];
```
zhuminjie, line 296: Same as above

SRC/recorder/ElementRecorder.cpp
```diff
@@ -247,8 +249,14 @@ OPS_ElementRecorder()
             nargrem = 1 + OPS_GetNumRemainingInputArgs();
             data = new const char *[nargrem];
             data[0] = option;
-            for (int i = 1; i < nargrem; i++)
-                data[i] = OPS_GetString();
+	    argv = new char*[nargrem];
```
zhuminjie, line 252: same as above

zhuminjie (changes requested):

fmckenna (approved):

fmckenna (approved):

PR #529, opened 2021-02-09
Update of LeadRubberX (Or LeadRubberX2)

shocky0424:
This element is a modification of the LeadRubberX element. The modifications are of importance in the analysis of lead-rubber bearings in long-duration ground motions with fluctuating intensity and in simulating behavior in experiments.

The attached file blow contains:

1. Demonstration of modified LeadRubberX element (say, LeadRubberX2)
2. 2 examples which compare the responses from LeadRubberX and LeadRubberX2 element under a) Scaled 2011 Tohoku Earthquake, b) Imposed harmonic motion
3. source files (LeadRubberX2.cpp and LeadRubberX2.h files)

[LeadRubber Submission.zip](https://github.com/OpenSees/OpenSees/files/5954688/LeadRubber.Submission.zip)

### DIFF
--- a/SRC/element/elastomericBearing/LeadRubberX.cpp
+++ b/SRC/element/elastomericBearing/LeadRubberX.cpp
@@ -20,7 +20,7 @@
 
 // Modification of LeadRubberX
 // Written: Manish Kumar (mkumar2@buffalo.edu)
-// Modified: Hyun-myung Kim (hkim59@buffalo.edu)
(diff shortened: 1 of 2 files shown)


shocky0424:
I intentionally changed the source file name from LeadRubberX2.cpp to LeadRubberX.cpp for showing the comparison.  Please let me know if I need other actions to do. (such as pull the requests again with the original file)

PR #536, opened 2021-02-16
Update of LeadRubberX (Or LeadRubberX2)

shocky0424:
This element is a modification of the LeadRubberX element. The modifications are of importance in the analysis of lead-rubber bearings in long-duration ground motions with fluctuating intensity and in simulating behavior in experiments.

The attached file blow contains:

1. Demonstration of modified LeadRubberX element (say, LeadRubberX2)
2. 2 examples which compare the responses from LeadRubberX and LeadRubberX2 element under a) Scaled 2011 Tohoku Earthquake, b) Imposed harmonic motion
3. source files (LeadRubberX2.cpp and LeadRubberX2.h files)

[LeadRubber Submission.zip](https://github.com/OpenSees/OpenSees/files/5990600/LeadRubber.Submission.zip)

### DIFF
--- a/SRC/element/elastomericBearing/LeadRubberX.cpp
+++ b/SRC/element/elastomericBearing/LeadRubberX.cpp
@@ -20,7 +20,7 @@
 
 // Modification of LeadRubberX
 // Written: Manish Kumar (mkumar2@buffalo.edu)
-// Modified: Hyun-myung Kim (hkim59@buffalo.edu)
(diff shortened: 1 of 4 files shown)


mhscott:
If LeadRubberX2 is a different formulation from LeadRubberX, then create a new class. You shouldn't change the class name just to make what appear to be a few minor modifications.

mhscott:
Also, please add the examples and documentation via a PR at the OpenSeesDocumentation repository:

PR #543, opened 2021-02-23
GetStringFromAll: if an input is a string, copy it to buffer

zhuminjie:
1. copy string
2. check int and double first


mhscott (approved):
Looks good!

PR #546, opened 2021-03-01
adding MVLEM_3D element

kkolozvari:
This PR includes addition of MVLEM_3D element, which is a 3D macroscopic model for nonlinear analysis of flexure-controlled RC walls. Please let me know if anything needs to be adjusted. Thank you!


kkolozvari:
User documentation and example files for the MVLEM-3D can be found here: [https://github.com/kkolozvari/MVLEM-3D](url)

kkolozvari:
User documentation and example files can be accessed from: [https://kkolozvari.github.io/MVLEM-3D/](url)

PR #549, opened 2021-03-02
Changes regarding RockingBC

vavgen:
I deleted some unnecessary preprocessor directives from the .cpp file (but not really solved the problem).
The main thing I did is reimport my files into the VS project, because initially I did it rather manually through editing the project files. This may have caused problems with the unique identifiers or because there was no alphabetical folder sorting. I suspect the problem was due to VS, since even before these changes, the files compiled successfully after some random edit in the RockingBC.cpp file and performing a build again.
The project now compiles successfully, I hope it's not random again...


mhscott:
Thanks, worked for me!  How about you, @kkolozvari and @ambaker1?

kkolozvari:
Thank you @mhscott and @vavgen . It works for me as well!

vavgen:
@mhscott @kkolozvari I'm glad! Sorry for the trouble.

ambaker1:
Worked for me as well!

PR #561, opened 2021-04-03
Added ndm and ndf utility commands

ambaker1:
Utility commands for getting the input ndm and input ndf, as well as the nodal ndm and ndf. Syntax:
getNDM<$nodeTag>
getNDF<$nodeTag>
If the nodeTag argument is provided, it will return the nodal dimensions/degrees of freedom, or throw an error if the node does not exist.
If the nodeTag argument is not provided, it will return the input ndm/ndf, or the blank string if the model builder is undefined.


ambaker1:
Alternatively, the commands could be called getNDM and getNDF

mhscott:
Looks like there's a conflict now after merge of getClassType

ambaker1:
It should be fixed now.

mhscott:
Wait like 45 minutes then check again :)

ambaker1:
The merge had some CR/LF issues, apologies for the large # of "modified" lines.

PR #562, opened 2021-04-03
Added nodeDist, utility distance calculator

ambaker1:
Computes the distance between two nodes, or optionally the signed distance in a specific dimension.
Syntax:
nodeDist $iNode $jNode <$dim>
e.g.
node 1 0 0
node 2 3 4
nodeDist 1 2 -> 5
nodeDist 1 2 1 -> 3
nodeDist 1 2 2 -> 4
nodeDist 2 1 1 -> -3
nodeDist 2 1 2 -> -4

This function, while not providing any information that you couldn't get with nodeCoord, is a convenience in cases where the length of an element is needed (e.g. defining plastic hinge length as proportion of element length)


mhscott:
We'd like to keep OpenSees as light as possible. As you mentioned this can be done with nodeCoord.

ambaker1:
Sounds good

PR #564, opened 2021-04-05
Added better mixed MP command

ambaker1:
Added mixedEqualDOF MP constraint command.
The current command to make a mixed MP constraint has rather cumbersome syntax. I am proposing to make a new command with better syntax to replace the equalDOF_Mixed command (while keeping the old command for backwards compatibility)

Current mixed MP syntax:
equalDOF_Mixed $rNode $cNode $numDOF $rDOF1 $cDOF1 $rDOF2 $cDOF2 ... ...
Proposed:
mixedEqualDOF $rNode $cNode $rDOF1 $cDOF1 $rDOF2 $cDOF2 ... ...

The main difference is the elimination of the redundant $numDOF argument, which can simply be calculated with integer division when parsing the command. Additionally, I think the proposed command name is cleaner.

### DIFF
--- a/SRC/domain/constraints/MP_Constraint.cpp
+++ b/SRC/domain/constraints/MP_Constraint.cpp
@@ -93,6 +93,89 @@ int OPS_EqualDOF()
     return 0;
 }
 
+int OPS_MixedEqualDOF()
+{
+    // Check number of arguments
+    if (OPS_GetNumRemainingInputArgs() < 4 || OPS_GetNumRemainingInputArgs() % 2 == 1) {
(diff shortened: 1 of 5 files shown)


mhscott:
We can use even/odd logic when parsing the current command to get the same effect. Don't need a new command if removing an input is the only difference.

ambaker1:
Hmm. Good idea. Should I open a new pull request with modifications to the equalDOF_Mixed command as you suggested?

ambaker1:
The only issue I see with even-odd-parsing is that error handling and error messages may be a bit confusing if someone inputs incorrectly. Would it be "equalDOF_Mixed $rNode $cNode <$numDOFs> $rDOF1 $cDOF1 $rDOF2 $cDOF2 ..." ?

PR #565, opened 2021-04-06
Element load output commands

sewkokot:
This PR adds a few utility commands which will be helpful for OpenSeesPy plotting module.
For now ops_vis module requires specifying the element load information in the plotting commands which can be different from the actual element load data in the domain. The commands uses unique class tags for elements and element loads specified in `OpenSees/SRC/classTags.h `

The new output commands and examples of usage are:
```
nEles = ops.getNumElements()

# zero or one ele_tag only
ele_classtag = ops.getEleClassTags(1)
ele_classtags = ops.getEleClassTags()

ele_load_classtags = ops.getEleLoadClassTags(1)
ele_load_classtags_all_patterns = ops.getEleLoadClassTags()
# error if pattern_tag does not exist
# ele_load_classtags = ops.getEleLoadClassTags(3)

ele_load_tags = ops.getEleLoadTags(1)
ele_load_tags_all_patterns = ops.getEleLoadTags()

ele_load_data = ops.getEleLoadData(1)
ele_load_data_all_patterns = ops.getEleLoadData()
```

The example output would be:
```
nEles: 2
ele_classtag: 3
ele_classtags: [3, 3]
ele_load_classtags: [3, 12]
ele_load_classtags_all_patterns: [3, 12, 4]
ele_load_tags: [1, 2]
ele_load_tags_all_patterns: [1, 2, 2]
ele_load_data:
[-4.0, 0.0, -3.0, -7.0, 1.1, 1.9, 0.1, 0.9]
ele_load_data_all_patterns:
[-4.0, 0.0, -3.0, -7.0, 1.1, 1.9, 0.1, 0.9, -7.0, 0.0, 0.5]
```

Please review it and consider it for merging.

### DIFF
--- a/SRC/interpreter/OpenSeesCommands.h
+++ b/SRC/interpreter/OpenSeesCommands.h
@@ -282,6 +282,7 @@ int OPS_sensLambda();
 int OPS_sensSectionForce();
 int OPS_sensNodePressure();
 int OPS_getNumElements();
+int OPS_getEleClassTags();
 int OPS_getEleLoadClassTags();
 int OPS_getEleLoadTags();
 int OPS_getEleLoadData();
--- a/SRC/interpreter/OpenSeesCommands.h
+++ b/SRC/interpreter/OpenSeesCommands.h
@@ -281,6 +281,10 @@ int OPS_sensNodeAccel();
 int OPS_sensLambda();
 int OPS_sensSectionForce();
 int OPS_sensNodePressure();
+int OPS_getNumElements();
+int OPS_getEleLoadClassTags();
+int OPS_getEleLoadTags();
+int OPS_getEleLoadData();
 // Sensitivity:END /////////////////////////////////////////////
 
 /* OpenSeesMiscCommands.cpp */
--- a/SRC/interpreter/OpenSeesOutputCommands.cpp
+++ b/SRC/interpreter/OpenSeesOutputCommands.cpp
@@ -45,6 +45,7 @@ UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
 #include <DOF_Group.h>
 #include <Matrix.h>
 #include <LoadPattern.h>
+#include <LoadPatternIter.h>
 #include <FileStream.h>
 #include <ID.h>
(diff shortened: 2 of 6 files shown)


sewkokot:
The conflict should be fixed now.

mhscott:
@sewkokot Can you look into the conflict? Probably came up with another recent merge.

mhscott:
Also, instead of class tags, it will be good to use getEleType after we get it implemented in Python. https://github.com/OpenSees/OpenSees/pull/560

sewkokot:
This commit: https://github.com/OpenSees/OpenSees/commit/7576bb3f557b9fc831736c95d3bc4029dce8c767
creates a lot of mess as it would seem that the whole commands.cpp file is changed and it's difficult to find out what has been actually modified. I need some time to resolve the conflict.

sewkokot:
> Also, instead of class tags, it will be good to use getEleType after we get it implemented in Python. #560

I think that the getEleType (based on getClassType) would not give us a unique information, instead the class tags are more reliable as they are unique. Also in `classTags.h` we have all the tags in one place.

For example in `ElasticBeamWarping3d.h` we have the same 
`const char *getClassType(void) const {return "ElasticBeam3d";};`
as in `ElasticBeam3d.h`. 

We have to check if all elements have the getClassType properly specified.

mhscott:
Good point! I'll talk the classTags/eleType over with @fmckenna.

Also, I thought @ambaker1 fixed the CR/LF issue with tcl/commands.cpp. Maybe it wasn't merged? You may have to close this PR and start a new one??

ambaker1:
Hi @mhscott, I'm not entirely sure what happened with the CR/LF with tcl/commands.cpp. I worked on the files on Windows 10, so I assume that I changed the files to CR/LF. 
As far as eleType (classTags) goes, I also noticed that some elements do not have a getClassType method, for example the SSPquad element. So, perhaps some work is needed to ensure that all elements have unique class types.

sewkokot:
Ok, @mhscott, I close this PR and open a new one.

PR #567, opened 2021-04-07
add pc command

zhuminjie:
for manually create pressure constraints


mhscott:
pc -> pressureConstraint

PR #571, opened 2021-04-15
Commit strain commands for material/section model builders

ambaker1:
This is so that you don't have to call the "setStrain" function multiple times to commit the state (e.g. you are looking for some tolerance on the response, for example).

### DIFF
--- a/SRC/modelbuilder/tcl/TclPlaneStressMaterialTester.cpp
+++ b/SRC/modelbuilder/tcl/TclPlaneStressMaterialTester.cpp
@@ -60,6 +60,11 @@ int  TclPlaneStressMaterialTester_setStrainPlaneStressMaterial(ClientData client
 							       int argc,   
 							       TCL_Char **argv);
 
+int  TclPlaneStressMaterialTester_commitStrainPlaneStressMaterial(ClientData clientData,
+    Tcl_Interp* interp,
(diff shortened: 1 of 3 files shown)


mhscott:
Thanks for looking at these testers @ambaker1. We'll probably hold off on merging for a while, until after we figure out the best way to combine the Python and Tcl implementations of these functions and may other utility commands. For example, over in SRC/interpreter/ you'll see that the setStrain command calls commitState automatically.

ambaker1:
Interesting, I was not aware of the functions in SRC/interpreter/. The testers in SRC/modelbuilder/tcl/ have an optional argument for the number of "setStrain" calls until they commit the strain. E.g. "model UniaxialMaterialTest 5" requires 5 iterations before it commits. I like the option of being able to decide when to commit the strain. 

For example, I've used the section tester to implement a moment curvature analysis that doesn't require creating new nodes/elements, and thus can be conducted seamlessly within or after defining any model.

Maybe as an alternative, have a function for setting the trial strain? This could be called "setTrialStrain". "setStrain" would then be the combination of "setTrialStrain" and "commitStrain". Or, there could be a "-trial" option for "setStrain".

ambaker1:
No longer working with these test model builders. See PR #594

PR #574, opened 2021-04-23
A constraint for 3D multi-scale interaction between Beam-Column element and solid element

karl424:
BSMPC is used to interact beam-column element  and solid element in 3D.


mhscott:
Please add documentation of this new constraint at https://github.com/OpenSees/OpenSeesDocumentation

mhscott:
Can you provide theory reference and documentation? Also, how does this constraint differ from those implemented by @parduino?

PR #584, opened 2021-05-26
OpenSees display "lineWidth" option

ambaker1:
This PR returns the default line width back to 1, which was overridden to be 2 in the latest release (#580), and provides a command that allows the user to change the line width on the scripting side.

Command: lineWidth $width
$width: Width of lines in pixels (integer). Default 1


mhscott:
A command for this is unnecessary, you can get more control by adding lineWidth as an optional input to the display command

ambaker1:
Good call. I'll implement that.

mhscott:
This will always evaluate to false. I think you meant || not &&

image

ambaker1:
Should be fixed now.

ambaker1:
Just for the record - default line width was changed back to 2 with these commits: c44da3def471c04170d60d25164ba80a06752e8a 6bac073f14da2e9d1537aba4a7657cd8f82a0cba

PR #597, opened 2021-06-18
OpenSeesPy Include Directory for Python

ambaker1:
Changed python path to be more general (not users/mhscott). @mhscott I'm not sure if this was intended.


mhscott:
Not intentional, I guess I chose the non-default path when I installed Anaconda.

Thanks for changing that!

ambaker1:
Thought so! By the way - my Tcl and Python installation aren't in the expected locations either, but I get around it by creating a directory junction: https://www.howtogeek.com/howto/16226/complete-guide-to-symbolic-links-symlinks-on-windows-or-linux/

PR #601, opened 2021-06-21
Re-Adding mpco recorder into tcl commands

MassimoPetracca:
Dear Dr. @fmckenna ,
For some reason, the "**mpco**" recorder command has been commented out from the tcl recorder command in this commit: 21ae8c7.

This PR re-adds it back to the tcl recorder commands.

If it has been removed on purpose because it was giving issues with someone's else code, please let me know, and I will see how to sort it out.
If instead it was removed by accident, please consider re-adding it.


fmckenna:
with the new CMake build process for OpenSees we will be using conan to install some of the external packages .. there exists a 1.12.0 conan recipe and a 1.10.6 available (https://conan.io/center/hdf5)

MassimoPetracca:
@fmckenna I see, Ok so I will go for the 1.12.0

fmckenna:
how is 1.12 going .. i am building using cmake and conan but am waiting for an upgrade to mpco to test it

MassimoPetracca:
@fmckenna I'm doing it right now. I hope to complete it today.

mhscott:
@MassimoPetracca OK, just let us know when it's ready. @fmckenna is still working on the conan build, so no rush.

MassimoPetracca:
@fmckenna I'm done with this Pull Request.

The final solution is:
In the MPCORecorder.cpp I check for the **_HDF5** macro
```cpp
#ifndef _HDF5
#define MPCO_HDF5_LOADED_AT_RUNTIME
#endif // !_HDF5 
```
If we have the HDF5 libraries (1.12.0) I statically link to them, in the same way as the DRM. **So both the MPCORecorder and the DRM link to the same libraries.**
If instead the **_HDF5 is not defined**, the MPCORecorder can still be compiled because it will use the "**link at runtime**" approach. In this case, the user does not need to have the HDF5 libraries for compilation.
However, if a user wants to use the MPCORecorder he will have to have the HDF5 shared libraries on his/her machine.

There is only one thing I do not understand:
in commit https://github.com/OpenSees/OpenSees/commit/5cf7ce102a9697d9effd4ffb732eb958876192e0, I've seen that you added a barrier using the _HDF5 macro.
Is this really necessary? As I said above, if the _HDF5 is not defined, the MPCORecorder will not try to link at compile time, so there should be no problem with that.
Or did I misunderstand something?

fmckenna:
@MassimoPetracca 

I figure if they want MPCO they need to have hdf5 installed when they build it .. if it turns out later they want MPCO, they can install hdf5 and rebuild OpenSees .. one benefit in including is if HDF5 evolves the user won't have to get older versions to run an older version of OpenSees as it will be included in the .exe .. and I do have users who want older versions of the code as newer versions might not have current ele or mat if author wants removed or they modify code in such a way as does not suit user.

MassimoPetracca:
@fmckenna 
Ok, I see.
So the MPCORecorder code is already ok.
Let me add a new commit to this PR so that I will add a check with the _HDF5 macro also around the code in the Tcl model builder.

MassimoPetracca:
@fmckenna 
I'm done with this PR. Whenever you want you can merge it.
I just checked it also using the Cmake build.
However, for the Cmake build I had to change something because now it is not working in the case of **OPS_Use_HDF5=ON**
I'll do another PR for this soon

fmckenna:
Taking out that variable .. might as well use the ones find bring in

PR #594, opened 2021-06-15
Material testing commands

ambaker1:
This PR is intended to improve upon the uniaxial material testing commands added in OpenSeesPy, expanding their application to n-dimensional materials and sections, and implementing them in the Tcl version as well. The benefit to this method of material/section testing is that it does not interfere with the rest of the OpenSees domain, thus it can be incorporated directly into an analysis without affecting the model.

### Commands that define the material/section to be tested:

- testUniaxialMaterial $matTag
- testNDMaterial $matTag (new!)
- testSection $secTag (new!)

These commands must be called to define the material/section to be tested. $matTag is the integer tag of the uniaxial/ND material to be tested, and $secTag is the integer tag of the section to be tested.

### Commands for changing the state of the material/section:

- setStrain $strain1 $strain2 ...
- setTrialStrain $strain1 $strain2 ... (new!)
- commitStrain (new!)

For uniaxial materials and uniaxial sections, setStrain and setTrialStrain require only one input. For n-d materials and non-uniaxial sections, the number of inputs should correspond to the number of degrees-of-freedom of the material/section. The commands setTrialStrain and commitStrain were added to have greater control over when the state of the material/section is committed.

### Commands for getting state information from the test material/section:

- getStrain
- getStress
- getTangent
- getDampTangent (only for uniaxial materials)
- getResponse (new!)

These commands return a single value (for uniaxial materials) or a list of values corresponding to the number of degrees-of-freedom (for n-dimensional materials and sections). An exception to this is getTangent, which returns the flattened (row-major) material/section tangent matrix. The getResponse command works in the same fashion as element recorders, for example, "damage" can be used for the Fatigue uniaxial material.

### Notes/examples

Additionally, this PR fixes a bug with the previous uniaxial material testing commands: **_the tester modified the state of the source material rather than creating a copy to test_**. This could lead to unexpected results if the material testing commands are used and then the materials are applied to elements. Here is a MWE of the bug this addresses (OpenSeesPy):
```
# Import OpenSees
import openseespy.opensees as ops

# Create the materials for the example
ops.wipe()
ops.model("BasicBuilder","-ndm",1)
ops.uniaxialMaterial("Elastic",1,100.0)
ops.uniaxialMaterial("MinMax",2,1,"-max",0.01)

# Using uniaxial material tester, set the strain to exceed the max strain
print("uniaxial material test")
ops.testUniaxialMaterial(2)
ops.setStrain(0.02)
print(ops.getStress())
print(ops.getTangent())

[...]

PR #602, opened 2021-06-21
Revert "Material testing commands"

mhscott:
Reverts OpenSees/OpenSees#594

Gonna revert this for now because it breaks the Linux build and there's no easy solution. @fmckenna @zhuminjie and I will sort it out after we get the Tcl/Python commands unified.


ambaker1:
Any update on when this will be sorted out?

mhscott:
Yes, after cmake and the OPS commands are resolved.

PR #611, opened 2021-07-02
IMKBilin minor fix

amaelkady:
A fix for an observed zigzag bug in the response during high frequency un/reloading.


mhscott:
Why is Mi_temp a private class variable when it's only used as a temporary variable in setTrialStrain of the .cpp file?

You can just declare a local variable
double Mi_temp = Mi_1 + K_j * (Ri - Ri_1);

It looks like there's a lot of private variables, perhaps with the same intent as only local variables. Consider using local variables where appropriate.

PR #616, opened 2021-07-09
Add an new element InertiaTruss to OpenSees

Xiaodong-Ji-THU:
The new element InertiaTruss is developed to simulate the behavior of inerter devices. Inerter is a two-terminal element which can generate inertial force proportional to the relative acceleration between its two ends. This new easy-to-use element was developed based on Truss element, and it was verified by comparison with Matlab simulation results and experimental test data. Use of this new element in simulating the interter-based dampers can be found in the paper (Ji X, Cheng Y, Molina Hutt C. Seismic response of a tuned viscous mass damper (TVMD) coupled wall system. Engineering Structures 2020; 225: 111252).

### DIFF
--- a/SRC/actor/objectBroker/FEM_ObjectBrokerAllClasses.cpp
+++ b/SRC/actor/objectBroker/FEM_ObjectBrokerAllClasses.cpp
@@ -138,7 +138,6 @@
 //#include "FiberSection.h"
 #include "FiberSection2d.h"
 #include "FiberSection3d.h"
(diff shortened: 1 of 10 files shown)


mhscott:
The new files revert other changes to OpenSees (look at the diffs). You always have to keep your code merged with upstream. Working offline then uploading your files is not OK.

mhscott:
https://gist.github.com/CristinaSolana/1885435

PR #619, opened 2021-07-12
add a new InertiaTruss element

Xiaodong-Ji-THU:
The new element InertiaTruss is developed to simulate the behavior of inerter devices. Inerter is a two-terminal element which can generate inertial force proportional to the relative acceleration between its two ends. This new easy-to-use element was developed based on Truss element, and it was verified by comparison with Matlab simulation results and experimental test data. Use of this new element in simulating the interter-based dampers can be found in the paper (Ji X, Cheng Y, Molina Hutt C. Seismic response of a tuned viscous mass damper (TVMD) coupled wall system. Engineering Structures 2020; 225: 111252).


Xiaodong-Ji-THU:
> 
> 
> Please add documentation at https://github.com/OpenSees/OpenSeesDocumentation
> 
> The element SRC/twoNodeLink/Inerter.cpp has the same functionality as your new element, plus additional features.

Thank you very much for your kind response.
We have added documentation and an example in OpenSeesDocumentation.
The InertiaTruss element is developed based on Truss element, while the Inerter element is based on twoNodeLink element. We have conducted case-study analysis to compare the two elements. The results of the case-study analysis indicates that both elements yield identical dynamic responses. In addition, the efficiency (computing time cost) of two elements is similar. Therefore, either one would be an option for simulating the inerters.

mhscott:
There is now a conflict for this PR.

Xiaodong-Ji-THU:
> 
> 
> There is now a conflict for this PR.

Dear Scott:
A merge overwrites our code in SRC/element/TclElementCommands.cpp line 155.
We have modified our code and solved this conflict.

PR #625, opened 2021-07-22
Update toOpenSeesPy.py

BijanSeif:
Hi master, Professor Scott

I have added lines 8 and 12 in the function body and it is working on the system and made it easier for beginners. So the usage if the function becomes more simple like line 9 and 10. I hope I have no mistake on this. 

With Best Regards


mhscott:
Consider using optional keyword arguments or a separate function. This change will break the current implementation.

PR #674, opened 2021-10-06
BugFix: About to hit the limit of number of if-elseif blocks!

MassimoPetracca:
@fmckenna @mhscott This is new to me!

I was recently working on new material models. Among them, there was a uniaxial material.
While I was adding the relevant code in the Tcl model builder, I followed the common rule of using the 
`
...
} else if (name is....) {
 // build my material
}
`

Then in VS, I found an internal compiler error due to the excessive number of consecutive else if blocks.
Googling I found out that the standard does not impose such a limit, but basically, all compilers have one.
It's simply very hard to find such a limit, because more than 256 else if blocks make the code unreadable.

In order to avoid this error, and to keep the number of changes to the minimum possible, I spit the else-if chain at the end using a pattern like this:
```
if(theMaterial == 0) {
 if (name is ...) {
   theMaterial = build my material...
 }
}
```

Of course, this will solve the problem only now, and only if other contributors will "append" their materials after this "brakepoint".
(The optimal solution would be a **complete refactoring** following the idea of the "map" used in the python version)

What do you think? Is this enough for now?


fmckenna:
i agree the tClmodelbuilder should use a map .. something i was planning on doing .. that was why we put it in the code the interpreter is using ..  holding off on putting into TclCOmmand file as that file will soon be using same one python does .. just have tp make sure all ele are in the other

PR #719, opened 2021-11-16
Fixing cmake build + handle the _HDF5 case

MassimoPetracca:
@fmckenna 

I was trying the cmake build on ubuntu and I found some compilation errors:

- in CMakeList.txt I had to comment lines such as:
`add_subdirectory("${PROJECT_SOURCE_DIR}/OTHER/SuperLU_5.1.1")`
for dependencies already included via the `opensees_load()` function in _BuildTools/cmake/OpenSeesDependenciesUnix.cmake_ to avoid errors like
```
  The binary directory
    /home/runner/work/OpenSees/OpenSees/build/OTHER/CSPARSE
  is already used to build a source directory.  It cannot be used to build
  source directory
    /home/runner/work/OpenSees/OpenSees/OTHER/CSPARSE
  Specify a unique binary directory name.
```

- in _BuildTools/cmake/OpenSeesDependenciesUnix.cmake_  I had to add the tetgen and triangle libraries to solve a **LINK Error**:
`opensees_load(tet BUNDLED "${OPS_BUNDLED_DIR}/Tetgen/")`
`opensees_load(triangle BUNDLED "${OPS_BUNDLED_DIR}/Triangle/")`
Probably the same should be done in the other **OpenSeesDependencies\*\*\*.cmake** files, but I'm not sure about them so I leave it to you...

- I had to comment this line in CMakeList.txt:
`target_include_directories(OpenSees PUBLIC ${TCL_INCLUDE_PATH})` to avoid this error:
```
  CMake Error at CMakeLists.txt:353 (target_include_directories):
  Cannot specify include directories for target "OpenSees" which is not built
  by this project.
```
- Finally, I had to fix some stuff related to the **OPS_Use_HDF5** options + OPS_DRM + OPS_ASDEA (MPCORecorder)

It compiles fine both with the default settings and with -DOPS_Use_DRM=ON  -DOPS_Use_HDF5=ON

**The only problem now is that you have to change the command you use to automatically run the cmake compilation as a check in GitHub, as it is still using the target OpenSeesTcl... I've seen that recently it was changed to OpenSees.**


fmckenna:
@MassimoPetracca 

I was fixing the UNIX one and the OpenSees/CMakeLists.txt .. i am had to do it a little differently as MPCO requires version >= 1.12.0 (or does it?) and DRM does not. The latest CMakeFiles work for mac with conan (which include the HDF5 1.12.0 recipe), linux with/without conan and with HDF5 version 1.10.8 (what came when i used apt install)

MassimoPetracca:
@fmckenna actually, when the **_HDF5** macro is defined (add_compile_definitions(_HDF5) if Ops_Use_HDF5=ON) the MPCORecorder will link statically and so **any version >= 1.10.0 will work** ( I need a function that was added starting from 1.10.0).

I tested it on Ubuntu 18.04.6 with the default package (libhdf5-serial-dev, whose version is 1.10.0) and it worked.
I also tested it with the 1.12.0 (I thought you need it for conan), manually installing it and specifying its location manually as the FindHDF5 recipe won't work in that case:
```
develop_dir=$HOME/Develop
packages_dir=$develop_dir/Packages
hdf5_dir=$packages_dir/hdf5/1.12.0/linux64
zlib_dir=$packages_dir/zlib/1.2.9/linux64

cmake .. \
-DCMAKE_BUILD_TYPE="Release" \
-DOPS_Use_DRM=ON \
-DOPS_Use_Dev_Directories=OFF \
-DOPS_Use_PFEM=ON \
-DOPS_Use_HDF5=ON \
-DHDF5_C_INCLUDE_DIR="$hdf5_dir/include" \
-DHDF5_hdf5_LIBRARY_RELEASE="$hdf5_dir/lib/static/libhdf5.a;$zlib_dir/libz.so.1.2.9"
```

Previously, when I was using the "link at runtime" I needed to re-define some variables (like the HDF5 version) inside the MPCORecorder.cpp as  I did not have access to them at compile time. But now this problem should not be relevant anymore,
so if even version 1.10.8 is fine.

PR #721, opened 2021-11-23
bugfix setResponse method in SectionAggregator

MassimoPetracca:
@fmckenna @mhscott 

This PR

1. Fixes a bug in the setResponse method of the SectionAggregator which can result in a segmentation fault if one asks for "section" output when the sections inside the Aggregator is NULL.
2. Due to a recent clean-up in the SectionAggregator, if one wants to ask for the fiberSection response inside the aggregator, should use "_section $sec_tag **section** fiber $fiber_tag stress_" instead of "_section $sec_tag fiber $fiber_tag stress_". The MPCORecorder has been updated accordingly


mhscott:
Thanks, @MassimoPetracca!

Issue #498, opened 2021-01-12
Steady-state response to harmonic load (quasi-static problem)


sewkokot:
I'm interested in solving the steady-state harmonic response due to
the steady-state harmonic load for an undamped model.

The load is as follows
F(t) = F0 * sin(p * t)
where 'p' is a circular frequency (rad/s) of the harmonic load, and F0 is a global amplitude vector
of nodal loads.

Therefore starting from the following equation of motion

M a(t) + K u(t) = F0 * sin(p * t)

this equation can be simplified to the system of algebraic equations

Since the response is harmonic
u(t) = u0 * sin(p * t) and
a(t) = -p^2 * u0 * sin(p * t)

the steady-state response comes from the following
quasi-static equation

(K - p^2 M) u0 = F0   or simply written as  
Kmod * u = F0        (as in statics Ku = F)

The F0 amplitude load distribution can be taken from the
loadPattern/load/eleLoad commands (timeSeries can be neglected here), 
only the excitation frequency 'p' should be specified somewhere.

Is there an easy way to implement it? Any hints on what classes
(integrator?) should be modified or a new command should be added similar to 'eigen'?

Thank you in advance for any help.

mhscott:
Hello Seweryn,

You can solve the original equation of motion M a(t) + K u(t) = F0 * sin(p * t) with a 'Sine' timeSeries and nodal load on an SDF system, or with a uniformExcitation.

Michael

sewkokot:
Thank you, Michael, however I need the steady-state response of a MDOF system and I want to avoid the transient analysis. I have a test example using the 'Trig/Sine' timeSeries with rayleigh damping to get rid of the transient phase, but it's time consuming. I need to do thousands of repetitions for various parameters so the direct quasi-static solution would be optimal.

A workaround would be to use your GimmeMCK integrator commands to get the K and M matrices and finally solve for the amplitudes using numpy's solve function:
u0 = np.linalg.solve(K - (p*p)*M, F0)

It seems that the standalone 'steadyStateResponse' command can be implemented natively in OpenSees, similarly to responseSpectrum from the following PR
https://github.com/OpenSees/OpenSees/pull/485

I will look at the responseSpectrum implementation to figure out the solution for the steady-state response. Any hints on how to implement it in OpenSees would be appreciated :)

mhscott:
[...]

PR #723, opened 2021-11-27
Adding HarmonicSteadyState integrator

sewkokot:
This PR implements the harmonic steady-state integrator as discussed in https://github.com/OpenSees/OpenSees/issues/498

The implementation is based on the LoadControl integration as a subclass of the StaticIntegrator, therfore the actual syntax of the 'HarmonicSteadyState' or 'HarmonicSS' is similar to the LoadControl integrator:

```
...
f = 148  # Hz
p = 2 * np.pi * f  # rad/s
# ops.integrator('LoadControl', 1.0)
ops.integrator('HarmonicSteadyState', 1.0, p)
```

where 'p' is the circular frequency in rad/s of the harmonic load F(t) = F0 * sin(p*t).

The following equation of motion
`
M a(t) + K u(t) = F0 * sin(p * t)`

is solved as a quasi-static problem

`(K - p^2 M) u0 = F0` or simply written as
`Kmod * u = F0` (as in statics `Ku = F`)

The F0 amplitude load distribution can be taken from the
loadPattern/load/eleLoad commands.

### DIFF
--- a/DEVELOPER/core/classTags.h
+++ b/DEVELOPER/core/classTags.h
@@ -910,6 +910,9 @@
 #define INTEGRATOR_TAGS_ExplicitDifference              55
 #define INTEGRATOR_TAGS_EQPath                          56
 #define INTEGRATOR_TAGS_GimmeMCK                        57
+#define INTEGRATOR_TAGS_StagedLoadControl               58
+#define INTEGRATOR_TAGS_StagedNewmark                   59
+#define INTEGRATOR_TAGS_HarmonicSteadyState             60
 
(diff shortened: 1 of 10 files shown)


sewkokot:
[Update]
I have added a new commit to make the input parameter consistent with 'Trig' time series, where the period Tp [in seconds] is used instead of the circular frequency p [in rad/s].

So now we have
`[K - (2*pi/Tp)^2 M] u0 = F0`

And an example usage:
```
...
f = 148  # Hz
Tp = 1/f  # s
ops.integrator('HarmonicSteadyState', 1.0, Tp)
```

PR #729, opened 2021-12-13
Adding fixed/retained/constrained Nodes/DOFs output commands to OpenSeesPy

sewkokot:
This PR adds the fixedNodes, fixedDOFs, constrainedNodes, constrainedDOFs, retainedNodes, retainedDOFs output commands to OpenSeesPy. Similar Tcl commands have been added by @ambaker1 in https://github.com/OpenSees/OpenSees/pull/563


ambaker1:
I was thinking about these commands - they aren't really documented yet, and I think it may be better to have "get" as a prefix for all of them. What do you think, @sewkokot ? It would be getFixedDOFs, getFixedNodes, etc.

sewkokot:
@ambaker1 I'm going to use these commands in the OpenSeesPy opsvis module and for the naming convention, indeed, the 'get' prefix would make them more consistent with other similar commands. However let's wait for @fmckenna and @mhscott to decide.

ambaker1:
@sewkokot Awesome! It would be cool for opsvis to show constraints. By the way, since creating these commands, I have realized that the way I implemented the MP constraint queries are a bit lacking, and I have a better idea for them. If you are interested, I'd love to collaborate on it.

sewkokot:
@ambaker1 Yes, I'm interested! Let me know what can be improved in these commands. Also if you have ideas how the constraints could be represented graphically (symbols, text?) in opsvis I would try to implement it. For now I have ported your Tcl commands into Py version to get the same output lists containing node tags and dofs.

ambaker1:
@sewkokot 
Here's a few of my ideas: 

1. The getRetainedDOFs and getConstrainedDOFs currently have an option to query of degrees of freedom retained or constrained from a specific degree-of-freedom. I'd like to remove that option, and instead add new command, say "getMPConstraintMatrix" that returns the constraint matrix for a given node pair. The way I implemented that optional feature before may be misleading in some cases.
2. Add an MP "type" property to MP constraints that could be queried with something like "getMPType". This would help with text display for your opsvis package.
3. Store MP constraint information in a more easily queried datastructure. The current implementation loops through all MP constraints for every query. Maybe a tree or dictionary structure?

sewkokot:
@ambaker1 I like your ideas! Below are my specific comments:

>     1. The getRetainedDOFs and getConstrainedDOFs currently have an option to query of degrees of freedom retained or constrained from a specific degree-of-freedom. I'd like to remove that option, and instead add new command, say "getMPConstraintMatrix" that returns the constraint matrix for a given node pair. The way I implemented that optional feature before may be misleading in some cases.

Indeed the constraint matrix would be better here. So let's implement such a command in Tcl and Python commands and fix the getRetainedDOFs and getConstrainedDOFs ones.

>     2. Add an MP "type" property to MP constraints that could be queried with something like "getMPType". This would help with text display for your opsvis package.

If the MP constraint type cannot be derived from the values of the constraint matrix, the additional MP type would definitely help. I need to check it on some test examples.


>     3. Store MP constraint information in a more easily queried datastructure. The current implementation loops through all MP constraints for every query. Maybe a tree or dictionary structure?

The problem is that, to my understanding, both Tcl and Python output commands can only return a list, so even the constraint matrix must be flattened out to a list as it is in the case of getting the mass, damping and stiffness matrices from the GimmeMCK/printA commands. @fmckenna, @mhscott, @zhuminjie am I right?

sewkokot:
@ambaker1 Thank you for the clarification! 

Coming back to this PR, the constraint matrix can be implemented in another PR because it requires more work. For now, this PR ports the already existing: `fixedNodes`, `fixedDOFs`, `constrainedNodes`, `constrainedDOFs`, `retainedNodes`, `retainedDOFs` Tcl output commands into OpenSeesPy with the suggested prefix 'get'. So for both Tcl and Python, the commands would be named as follows: `getFixedNodes`, `getFixedDOFs`, `getConstrainedNodes`, `getConstrainedDOFs`, `getRetainedNodes` and `getRetainedDOFs`.

An optimization of these commands can also be done later as soon as the users need it.

In the current state, this PR would be useful in `opsvis` for plotting supports. Now the supports are plotted using `node_dofs = ops.nodeDOFs(node_tag)` which does not return any DOFs before running the `analyze` command. And using `getFixedNodes` and `getFixedDOFs` does not have this limitation.
See an opsvis example: https://opsvis.readthedocs.io/en/latest/ex_2d_portal_frame.html

@fmckenna, @mhscott Therefore I'm kindly suggesting for merging this PR and further improvements can be done in a separate PR? What do you think?

PR #744, opened 2021-12-26
fixing bug with internal nodes tag

kkolozvari:
tags of internal nodes are now based on element tag, not tag of the first node. this will help avoid issues with trying to create the same internal node from two different elements.

### DIFF
--- a/SRC/element/mvlem/SFI_MVLEM.cpp
+++ b/SRC/element/mvlem/SFI_MVLEM.cpp
@@ -578,12 +578,12 @@ void SFI_MVLEM::setDomain(Domain *theDomain)
     opserr << "WARNING: Element is NOT vertical!";
   }
   
-  
+  int eletag = this->getTag();
   // Create a internal node tag
   for (int i = 0; i < m; i++){ // Large NEGATIVE integer starting with tag of the element
-    externalNodes(i+2) = -(Nd1*1000 + i + 1); // Max fibers is 999 to avoid overlap
+    externalNodes(i+2) = -(eletag*1000 + i + 1); // Max fibers is 999 to avoid overlap
   } 
-  
+
   // Build m internal nodes (NodesX) and add them to the domain
   for (int i = 0; i < m; i++) {
     
--- a/SRC/element/mvlem/SFI_MVLEM_3D.cpp
+++ b/SRC/element/mvlem/SFI_MVLEM_3D.cpp
@@ -628,9 +628,10 @@ void SFI_MVLEM_3D::setDomain(Domain *theDomain)
 	// Calculate distance of corner nodes from center axis
 	d = Lw / 2.0;
 
+	int eletag = this->getTag();
 	// Create a internal node tag
 	for (int i = 0; i < m; i++) { // Large NEGATIVE integer starting with tag of the element
-		externalNodes(i + 4) = -(Nd1 * 1000 + i + 1); // Max fibers is 999 to avoid overlap
+		externalNodes(i + 4) = -(eletag * 1000 + i + 1); // Max fibers is 999 to avoid overlap
 	}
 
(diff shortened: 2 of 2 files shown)


mhscott:
Inevitably, using "arbitrary" tags for internal nodes will lead to errors. Users do unexpected things.

Why do you need to create nodes (1 DOF) and add them to the Domain in the first place? You can keep track of the displacements internally with an array/vector. There's no mass assigned to these nodes, correct?

kkolozvari:
Internal DOFs are part of element formulation. They are defined at each fiber of the element and correspond to horizontal expansion of the plane-stress material assigned to each fiber (FSAM material); vertical expansion and shear deformation are obtained from the element DOFs. These DOFs at each fiber are part of element stiffness matrix and force vector, which is why I decided to create internal element nodes. There might be another way to achieve the same thing, but that would require complete redesign of the element.

In terms of tags of these internal nodes - In the original 2D 2-node formulation, I decided to assign them as negative integers that are generated automatically based on a tag of the first node, assuming that users would not intentionally define nodes with large negative values (still think that is very unlikely). This formulation worked on for 2D 2-node element since you would almost never need to connect two SFI-MVLEM elements to the same node in a 2D problem. However, in 3D 4-node element formulation this is not the case. Therefore, I decided to fix the issue by generating tags of internal nodes based on element tags, which is unique. I agree with you that users can do unexpected things, but I think this fix should be good enough for vast majority of applications. I have used the model for years and never had any issues even before the fix was implemented. I would appreciate if you can accept the PR as it does reduce the risk of a big significantly.

By the way, what is the deal with Ubuntu check failing? Is there anything I need to do on my end to fix this? I just changed a couple of lines of the code. Please advise. Thank you.

fmckenna:
the internal node tags MUST be part of the element command as is done for example in the Joint2D and 3d elements .. that way no user who understands your element will run afoul of some well thought out but hidden node numbers .. 
 now something about smart node numbers could be put in documentation, but nobody reads the fine print of the documentation so it will never be a safe answer as if enough people are using your element someone will always run afoul of this problem.

kkolozvari:
@fmckenna thanks for the tip. Please note that 2D version of the model had the same implementation of the internal nodes and many OpenSees users have been successfully using it over the years. 

I will take a look at Joint2D/3D elements and work on improving the model, but this can take some time. In the meantime, I would really appreciate if you can approve this PR because, although not perfect, it reduces the chance of error to a very very small percentage. The way the model is implemented now, that chance is much higher and it is likely that others will run into the same issue. Therefore, this PR is still a significant improvement for what is currently in OpenSees. 

Thank you

fmckenna:
I will pull it in now .. i will give you a month to update .. after that i will make the modifications to the code.

kkolozvari:
@fmckenna thank you!

kkolozvari:
I looked at Join2D/3D code. It is doing essentially the same as SFI-MVLEM, except the user defines tags of internal nodes manually. In the case of SFI-MVLEM, this means that for each fiber of the wall model the user will need to define node tag manually. This will make use of this element extremely unpractical if this is the only option. 
I am thinking about implementing few options for definition of tags of internal nodes to make the use of the element user friendly and transparent in the same time:
Option 1) allow the user to manually define tag of each internal node as an array
Option 2) instead of user having to define each tag manually, have the user define tag only of the first internal node and other tags would be defined automatically in consecutive order. that way, the user is aware of internal node tags, but it doesn't have to define each tag
Option 3) add a switch so current version of the tags definition is still available and emphasize what the model is currently doing in the user documentation. I would really like to keep this option in the code.
In sum, there will be a new input switch for options 1, 2, 3 above so the user can choose which option to use for definition of internal nodes.

Please let me know what you think. Thank you

Issue #695, opened 2021-10-27
small issue in setResponse method of Shell sections


MassimoPetracca:
Dear @mhscott ,  in a recent pull request #630 all sections are now calling their base class implementation to avoid duplicated code.
However, the base class opens the XML tags for the component labels (example: output.tag("ResponseType","Mz");) in a switch, assuming the components are those of a beam element.

This is not a major issue, it's only relevant for recorders with the -xml flag.
Now all components of shell sections are "Unknown".

mhscott:
So, we just need to add more cases to the switch statement in the base SectionForceDeformation::setResponse implementation?

MassimoPetracca:
@mhscott Yes, we need to add extra switch statements. But in order for it to work, we also need to:

1. Declare more SECTION_RESPONSE_XXX macros in SectionForceDeformation.h
```
// membrane forces
SECTION_RESPONSE_FXX
SECTION_RESPONSE_FYY
SECTION_RESPONSE_FXY
// bending moments
SECTION_RESPONSE_MXX
SECTION_RESPONSE_MYY
SECTION_RESPONSE_MXY
// transverse shear
SECTION_RESPONSE_VXZ
SECTION_RESPONSE_VYZ
```
2. In all shell sections (ElasticMembranePlateSection, MembranePlateFiberSection, LayeredShellFiberSeciton and their thermal couterparts) , Implement the SectionForceDeformation::getType() -> ID method, which should return an ID with the above mentioned codes.

If you agree, I can create a PR for this during the weekend

mhscott:
Thanks, this looks good. Please make a PR when you are ready.

MassimoPetracca:
Dear @mhscott I just created a PR #755

PR #755, opened 2021-12-30
Fixing setResponse methods in all shell sections

MassimoPetracca:
Dear @mhscott , 
as discussed in this issue #695 , due to a recent clean-up of shell cross sections, the common responses such as stress and strain are handled by the base class SectionForceDeformation.
However, SectionForceDeformation recognizes only responses of beam elements, thus the XML tags of the components of shell elements are now all "UnknownStress".

This PR:

- Defines the following macros for responses of shell sections in SectionForceDeformation.h: 
```cpp
#define SECTION_RESPONSE_FXX 11 // membrane xx
#define SECTION_RESPONSE_FYY 12 // membrane yy
#define SECTION_RESPONSE_FXY 13 // membrane xy
#define SECTION_RESPONSE_MXX 14 // bending xx
#define SECTION_RESPONSE_MYY 15 // bending yy
#define SECTION_RESPONSE_MXY 16 // bending xy
#define SECTION_RESPONSE_VXZ 17 // bending yy
#define SECTION_RESPONSE_VYZ 18 // bending xy
```
- Implements the proper getType() methods in all shell sections to match those values
- The MembranePlateFiberSection now opens XML tags consistently with the LayeredShellSection for fiber response
- In the DoubleMembranePlateFiberSection there was a small bug ("or" instead of "and") in the if-statement for fiber results
- In the DoubleMembranePlateFiberSection the fiber output was given only for the first 5 fibers. Now all 10 fibers are handled

### DIFF
--- a/SRC/material/section/DoubleMembranePlateFiberSection.cpp
+++ b/SRC/material/section/DoubleMembranePlateFiberSection.cpp
@@ -846,29 +846,33 @@ DoubleMembranePlateFiberSection::setResponse(const char **argv, int argc,
 {
   Response *theResponse =0;
 
-  if (argc > 2 || strcmp(argv[0],"fiber") == 0) {
+  if (argc > 2 && (strcmp(argv[0], "fiber") == 0 || strcmp(argv[0], "Fiber") == 0)) {
     
     int passarg = 2;
     int key = atoi(argv[1]);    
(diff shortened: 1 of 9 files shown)


mhscott:
Thanks!

PR #761, opened 2022-01-16
Begin implementing threadsafe runtime handling.

claudioperez:
<center>
<i>Under development</i>
</center>

## Outline
1. **Runtime**
2. **commands**

## Introduced Macros

- `OPS_USE_RUNTIME`
- `OPS_DECL_RUNTIME(function, ...)`
- `OPS_CALL_RUNTIME(function, ...)`

### UW elements
Problem: Some UW elements call API functions (ie `OPS_getCrdTransf`) from the class constructor.

Solution: At this stage the runtime pointer is only being passed to element `OPS_*` constructors, meaning that these elements will not work when `OPS_USE_RUNTIME` is defined.

### Absorbent boundary elements
Problem: Some of these elements create closures (ie lambdas) with no default capture clause.
Solution: Add default capture clause.


SRC/element/forceBeamColumn/ElasticForceBeamColumnWarping2d.cpp
```diff
@@ -29,7 +29,7 @@
 Shear Wall with Warping
  ---
 Hsiang-Chuan Tsai, James M. Kelly (2004), "Buckling of short beams with warping effect included."
-International Journal of Solids and Structures, 42:239�253
+International Journal of Solids and Structures, 42:239�253
```
claudioperez, line 32: Note to self: Change this to something ASCII.

PR #763, opened 2022-01-19
Fixing Visual Studio Compilation due to DuctileFracture uniaxial material

MassimoPetracca:
@fmckenna @mhscott 
This PR fixes a compilation error in Visual Studio.
The new DuctileFracture uniaxial material was called in the TclModelBuilderUniaxialMaterialCommand.cpp inside an **else if** block which caused a compilation error due to an excessive number of **else if** blocks.

I simply moved it at the end using a separated **if** block


mhscott:
I don't have this issue on VS 2019. Are you using an older version?

MassimoPetracca:
@mhscott It's because I have a new uniaxial material under development and I hit this if-else limits. It will happen the next time someone adds a new uniaxial material with an "else if".
So the solution is to simply add it at the end within a new "if" (not "else if")

PR #766, opened 2022-01-21
Fix sprintf format in eleReponse (Tcl)

MassimoPetracca:
@fmckenna  @mhscott 
The "eleResponse" in the Tcl interpreter is the only command that uses "%f" as format in sprintf. This translates in numbers such as 1.0e-7 printed as 0.000000.
I changed it to the format used in all other commands "%35.20f"

### DIFF
--- a/SRC/tcl/commands.cpp
+++ b/SRC/tcl/commands.cpp
@@ -6606,7 +6606,7 @@ eleResponse(ClientData clientData, Tcl_Interp *interp, int argc, TCL_Char **argv
       int size = data->Size();
       char buffer[40];
       for (int i=0; i<size; i++) {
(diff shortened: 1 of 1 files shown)


mhscott:
Thanks! We really need to get rid of all the %35.20f stuff. Long story as to why it's there, but it's overkill and we need a more sensible format like %.16g or something

MassimoPetracca:
@mhscott I completely agree that the (f) fixed format is to avoid. I also would prefer the (g) generic format.
Do you want me to take care of it in this PR?
Or is it something that you already put in your todo list?

mhscott:
Thanks! We'll take care of it after checking that it doesn't break any old reliability code.

PR #772, opened 2022-02-07
Changes for compiling OpenSeesPyMP on Mac

millen1m:
This commit only contains changes required to compile 
OpenSeesPyMP (multi-processor version with Mumps) on Mac.
Notes:

- To use mumps need to use Metis5. Which causes a file name clash with metis.h file, therefore readnamed this to MetisWrapper.h/.cpp since this file is only a library wrapper.
- Had to do some minor mods of SUPERLU_5.1.1
- Included more steps in Make file.


claudioperez:
This broke the CMake build, but it looks like the fix will be quick; I'll open a PR shortly. We should also check the VS build @mhscott .
Also, the changes made to the Makefile here switch from using the bundled version of METIS; should this be a user-specific change? Is it possible to obtain a version of MUMPS that is compatible with the bundled METIS on Mac?

mhscott:
VS build is OK, no problems

millen1m:
Thanks for fixing this @claudioperez  - I haven't got my head round CMake yet. Though when I run cmake on mac it has further errors with the SuperLU package, everything else seems to build fine - though my understanding is that it is not the parallel version. Is there a way to change the build to parallel using mumps/metis5? 

errors with SuperLu with Cmake:

SuperLU_5.1.1/SRC/dsnode_bmod.c:103:2: error: 
      implicit declaration of function 'dlsolve' is invalid in C99
      [-Werror,-Wimplicit-function-declaration]
        dlsolve ( nsupr, nsupc, &lusup[luptr], &lusup[ufirst] );

and 

SuperLU_5.1.1/SRC/dsnode_bmod.c:104:2: error: 
      implicit declaration of function 'dmatvec' is invalid in C99
      [-Werror,-Wimplicit-function-declaration]
        dmatvec ( nsupr, nrow, nsupc, &lusup[luptr+nsupc],

PR #779, opened 2022-02-09
Fix MPIDiagonal

millen1m:
This commit only contains changes required to fix MPIDiagonal see attached test script - not committed.

 - One fix was a revert of list size made in commit de04944c7f09cab11a556a534c3194f0d7a04bc8

Not sure where to put such test scripts? Especially MP ones. Any advice would be helpful. Trying to systematically go through the fixes/changes that I have made in my branch and push the good changes to the main repo.

Test script fails
[mpidiagonal-error-message.txt](https://github.com/OpenSees/OpenSees/files/8028505/mpidiagonal-error-message.txt)
 with memory alloc error without these changes. See attached error.
[mpi_paralleltruss_explicit_py.txt](https://github.com/OpenSees/OpenSees/files/8028503/mpi_paralleltruss_explicit_py.txt)


mhscott:
Can you explain more on why the extra count++ on line 141

millen1m:
Error on my part when I was doing some debugging and forgot to remove. Good spot.

PR #780, opened 2022-02-09
Added Variable transient analysis option to OpenSeesPy.

millen1m:
Pretty sure this is a bug, my compiler always tells me to fix it. and the header only defines DistributedSparseGenRowLinSOE. I guess if it isn't a bug then I will include a comment to avoid others also trying to fix it.


millen1m:
Sorry forgot to create a new branch for this extra commit - and having all sorts of weird issues with trying to rebase.
Anyway, since the commit above is so minor I will continue on this same one. If there is a problem I will re-attempt the rebase.

Variable transient wasn't connected, I have included it with minimal changes similar to the Tcl implementation. Initially I fixed this by changing the base analyze command to accept 5 inputs, but then you can't overload this function - the current implementation allows you to run both variable and normal transient analysis without error - which is quite useful.

I also noted that it seems that in the Tcl version 'theTransientAnalysis' is global and getTransientAnalysis() doesn't need to be called everytime analyze() is envoked. Not sure how to get theTransientAnalysis to be global or how much this impacts performance but could speed things up a tad.

PR #789, opened 2022-02-17
setParameters in ElasticMembranePlateSection

MassimoPetracca:
@mhscott @fmckenna 
I added the setParameter function to the ElasticMembranePlateSection.

I also added a setResponse for those parameters. It can be useful for updating a parameter based on the current value it has.


mhscott:
Why are you using getResponse/setResponse to update parameters when you could use setParameter/updateParameter?

MassimoPetracca:
Yes, I added the setParameter/updateParameter pair to handle the updating of parameters.
The setResponse/getResponse is not strictly necessary for the updating of parameters, but it can be useful when you want to programmatically update parameters as a function of their current value.

A typical scenario where I want to randomly apply, in a parametric study, a "damaged stiffness" on some elements:
```
damage = paramter value (from 0 to 1)
foreach element
     E0 = eleResponse(element, "E")
     E = (1-damage) * E
     setParameter(E, element, "E")
```

MassimoPetracca:
_"Use the getParamValue command"_
_"Pass the actual parameter values here, don't make them 0.0"_
Whoa... I was not aware of this function, thanks! I'll change it right now and do a commit soon

mhscott:
I have used it for many years for finite difference calculations and there's many other potential uses
https://portwooddigital.com/2021/11/27/finite-differences/

MassimoPetracca:
Great post!
I just updated the code, thanks for your suggestion!

PR #794, opened 2022-02-20
Changes for Cmake on MacOS

millen1m:
These changes were required to compile OpenSees on macos with cmake. There were two seperate issues:

 1. Reliability was defined outside of if statement - now in if statement see OpenSeesParameterCommands.cpp line 51. 
 2. Issues with SuperLu. missing definitions of functions.

Note sure if changes in 2. are necessary.


pletzer:
Linux, yes I have the latest version on master (just cloned).

PR #795, opened 2022-02-25
Added OpenSeesPy to CMake build process

millen1m:
This adds only 10 seconds to the build process and allows the OpenSeesPy package to be checked.

Note: I can't get it to build the parallel version. Locally I compile with mpicc and mpicxx but these are not allowed in the conan compiler settings and I don't know how to add them.


mhscott:
Examples are now in https://github.com/OpenSees/OpenSees/tree/master/EXAMPLES/ExamplePython

PR #810, opened 2022-03-17
add warning for eigen fullGenLapack solver

zhuminjie:


mhscott (approved):

BijanSeif:
Hi

According an investigation on the speed of 'fullGenLapack' and default eigen solver maybe it is better to say:

"WARNING - the 'fullGenLapack' eigen solver is VERY SLOW **if you want less than half of total eigen values**. Consider using the default eigen solver."

It has been seen that even for medium numbers of eigen values sometimes 'fullGenLapack' is faster than default eigen solver. (We are speaking about huge models with large stiffness and mass matrix)

PR #812, opened 2022-03-18
Add uniaxialMaterial SPSW02 to OpenSeesPy

ChuandongXie:
SPSW02 was developed by S.A. Jalali and have been available in OpenSees, but not available in OpenSeesPy yet.


sa-jalali:
Thank you @ChuandongXie for enabling Python invocation of the SPSW02 material.
Regards
SA Jalali

ChuandongXie:
> 

Thank you @sa-jalali for developing the SPSW material. These days I'm working on SPSW, and it seems the imperfection brings a great degradation in tension stiffness when plates have a large height-thickness ratio (600 and 900 in my study). I've already committed a code incorporated with imperfection. If you don't mind you can make a review and we can put it in SPSW02, or I commit a new one based on SPSW02.

PR #819, opened 2022-03-24
Compatibility improvements for Tcl

ambaker1:
The Tcl load command is now renamed to "oldload" instead of "import". The "import" command has special meaning for some Tcl packages.
https://core.tcl-lang.org/tcllib/doc/tcllib-1-20/embedded/md/index.md#import

Additionally, to better support compatibility with the Tcl load command, the "wipe" command now restores the Tcl load command, and an alternate alias, "nodeLoad" is provided.

This maintains backwards compatibility, and allows for binary files (e.g. some Tcl packages) to be loaded in between OpenSees analyses.

### DIFF
--- a/SRC/modelbuilder/tcl/TclModelBuilder.cpp
+++ b/SRC/modelbuilder/tcl/TclModelBuilder.cpp
@@ -513,6 +513,9 @@ TclModelBuilder::TclModelBuilder(Domain &theDomain, Tcl_Interp *interp, int NDM,
   Tcl_CreateCommand(interp, "load", TclCommand_addNodalLoad,
 		    (ClientData)NULL, NULL);
 
+  Tcl_CreateCommand(interp, "nodeLoad", TclCommand_addNodalLoad,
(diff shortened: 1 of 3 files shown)


claudioperez:
Hey this is a great catch. How about instead of `oldload`, we namespace it with something like `lib::load` or `lib::import` ? `oldload` might give the impression that its deprecated or superseded. It might even be possible to access the old command using `::load`, I can try running some test.

ambaker1:
@claudioperez 
Thanks!
I chose “oldload” to be consistent with how the “puts” command is treated. “Puts” in OpenSees actually is a redefined version of the Tcl “puts” command, which can be accessed via “oldputs”. 
I do now see that while the “puts” command is redefined with the same functionality, load is redefined with a completely different functionality. So renaming it to “oldload” may be confusing, even though it is just used as a placeholder to be renamed back to load after “wipe”. 
What if it was changed to “tclload” instead?

Also, ::load will be the same as load, if in the global environment. You can test it out yourself, but I am confident that you cannot access the original load command that way.

PR #845, opened 2022-05-17
-adding Steel05 uniaxialMaterial

sa-jalali:
Dear administrates (@mhscott, @fmckenna),
This pull request adds a new steel model (named as Steel05) which proposes a modification to the Monegotto-Pintho steel (Steel02) material. The modification includes adding in-cycle and cyclic deterioration capabilities, while keeping all "Steel02" behavior. The in-cycle degradation can take the place of joining Steel02 with MinMax material which posed numerical instabilities due to abrupt change in material's stiffness and strength. The cyclic degradation is mainly intended for uses other than the stress-strain model required in the fiber cross-sections, since for such uses the Menegotto-Pintho model already provides some source of cyclic degradation.
For the in-cycle (monotonic) degradation, a descending branch is added via a smooth transition curve that enhances convergence. The starting strain and the slope of the branch are taken from the user.
For the cyclic deterioration, I have used the hysteretic energy formerly used in Ibarra-Median-Krawinkler model. This model accepts a factor which is multiplied by the (Sigma_y*epsilon_y) to compute the energy absorption capacity, in addition to the c exponent used for the cycles' energy.
Further description and illustrative examples will be added to the documentation repo as soon as it becomes ready.
 
Actions taken:
-updating visual studio project
-vs build: successful
-updating SRC/makefile
-make: not tested

Thanks in advance for considering this pull request.

### DIFF
--- a/SRC/Makefile
+++ b/SRC/Makefile
@@ -881,6 +881,7 @@ MATERIAL_LIBS   =  $(FE)/material/Material.o \
 	$(FE)/material/uniaxial/SteelFractureDI.o \
 	$(FE)/material/uniaxial/DowelType.o \
 	$(FE)/material/uniaxial/DuctileFracture.o \
+	$(FE)/material/uniaxial/Steel05.o \
 	$(FE)/material/nD/NDMaterial.o \
 	$(FE)/material/nD/PlaneStressLayeredMaterial.o \
 	$(FE)/material/nD/PlaneStressRebarMaterial.o \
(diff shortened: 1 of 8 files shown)


sa-jalali:
There was an error in CMake Build related to my code which I addressed and made a new commit. The current error is related with an unresolved constructor method of CrdTransfResponse class.

fmckenna:
can you provide documentation.. please place in OpenSeesDocumentation repo.

PR #848, opened 2022-05-25
Adding New Material Models

msalehi2004:
Hello,

I would like to add a number of material models recently developed by me, Prof. Petros Sideris (TAMU), and Prof. Reginald DesRoches (Rice) to the public OpenSees. The descriptions of these models are found in this [journal paper](https://www.sciencedirect.com/science/article/pii/S0141029621018721). The models "Remove", "Delay", and "Switch" are time-dependent and could be used to apply changes to material models over time, for example, to simulate the repair of RC members. The models "BarBuckling" and "BarSlip2" could be used to simulate bar buckling and bar slip, respectively. All these material models are of "wrapper" type and "wrap" around other material models. The time-dependent models use a global variable ops_t. All these models have been tested using models defined in tcl (not Python) and several validation examples can be found in the above paper. The material model command descriptions will be added to the command manual once the codes are pulled into the master repository.

Please let me know if there are any issues or if you need further information before accepting the merge.

Thank you,
Mohammad


mhscott:
Looks interesting! How much of this could be accomplished at the script level, e.g., delay, switch, and remove based on using the getTime command? What is the reason for making extra classes?

msalehi2004:
Hello Michael,

These material models would let you make changes to the fiber sections in fiber elements or the materials in any element formulation during a time-history analysis. Even though the getTime command could give you the current time at the script level, one could not make changes to the materials in a predefined fiber section or element during a time-history analysis at the script level. As far as I know, once a fiber section or an element is defined in a model, it cannot be changed during analysis. As a result, any desired time-dependent changes have to happen within the material model/element classes.

Thanks,
Mohammad

mhscott:
Can you put the updates to GradientInelasticBeamColumn in a new branch then make a separate pull request?

mhscott:
Here is a video on using branches in OpenSees: https://youtu.be/kYMHQrJNIm0

mhscott:
No. I'm talking about another *branch* not another fork.

msalehi2004:
My apologies! I just did what you asked.

BTW, could you please let me know if there are any problems with this pull request that need to be resolved by me before you accept it?

Thank you!

mhscott:
You need to remove the GI files from this PR.
Also, you'll need to add documentation at https://github.com/OpenSees/OpenSeesDocumentation

fmckenna:
I notice you created a new global variable for the current time .. there is a global variable to the domain and this info can be obtained from there .. virtual double  getCurrentTime(void) const; .. we do not like adding global variables as this will break a lot of applications that build their own application .. you need to change this request to remove adding the global variable.

PR #853, opened 2022-05-29
Add the stress response for the truss element

sewkokot:
This PR add the stress response for the truss element. In many optimization problems for trusses, the constraints are expressed in terms of  stresses and therefore being able to get stresses directly would be very useful.


mhscott:
You can get stress from the material response:
eleResponse(eleTag,'material',1,'stress')

sewkokot:
@mhscott Thank you very much! Indeed it works, so I close this PR.

PR #858, opened 2022-05-30
Mvlem

mhscott:
Cleaning up some Vector references when using setResponse and fixing memory leak in SFI_MVLEM_3D


mhscott:
@kkolozvari I fixed some memory leaks in the MVLEM elements and FSAM material. Please test to make sure your examples/verifications are unchanged

PR #865, opened 2022-06-03
Defining fiberData options for recorder

mhscott:


BijanSeif:
Thanks @mhscott for this update. As I understood still "fiberData" returns same data as previous and to ask for new level of data we should use "fiberData2" option. Am I correct about the usage?

mhscott:
Yes, I was keeping things consistent with what's in the uniaxial fiber sections. fibeData is the original option and fiberData2 includes the material tag.

PR #871, opened 2022-06-06
Update QuadPatch.cpp

BijanSeif:
Hi Prof. @mhscott 
This update is for QuadPath to always get Positive Areas and modify clockwise definition and make them counter clockwise. Because I'm not perfect in C++ please check if did I use correct #include (lines 35 - 37) and also is definition of vertexCoordtmp is correct?
The algorithm checked in DevC++ by various samples.

### DIFF
--- a/SRC/material/section/repres/patch/QuadPatch.cpp
+++ b/SRC/material/section/repres/patch/QuadPatch.cpp
@@ -32,6 +32,9 @@
 #include <QuadPatch.h>
 #include <QuadCell.h>
 #include <elementAPI.h>
+#include <math.h>       
+#include <iostream>
+using namespace std;
 
 void* OPS_QuadPatch()
 {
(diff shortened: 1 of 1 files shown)


BijanSeif:
Hi again Prof. @mhscott 
Option  '-positiveArea' was added to the code. That would be great if code be checked. Thanks.

mhscott:
@BijanSeif I pulled your branch and the code did not compile. I corrected the compiling errors, but the fibers were still negative after using the '-positiveArea' options. Please compile locally and provide a minimal test to check that the algorithm is correct. I think all you have to do is switch the J and L vertices and then swap the num fibers in the IJ and JK directions.

BijanSeif:
Really thanks prof. @mhscott . I will fix it and check it locally then I will pull it.

PR #881, opened 2022-07-01
GI Element's Python Command Update

msalehi2004:
Hello Michael,

Here I have only made changes to the GI element commands.

Thanks,
Mohammad

### DIFF
--- a/SRC/element/gradientInelasticBeamColumn/GradientInelasticBeamColumn2d.cpp
+++ b/SRC/element/gradientInelasticBeamColumn/GradientInelasticBeamColumn2d.cpp
@@ -174,7 +174,7 @@ void* OPS_GradientInelasticBeamColumn2d()
 	const ID& secTags = theRule->getSectionTags();
 	int numIntegrPoints = secTags.Size();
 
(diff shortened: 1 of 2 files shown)


mhscott:
Your changes revert changes made earlier today. https://github.com/OpenSees/OpenSees/commit/60d0e79d6b2717cefd201715ff2d22b101071694
Please make sure you are always working with the latest code, i.e., pull from upstream, merge branches, etc.

msalehi2004:
Sorry, I wasn't aware of those changes. I just got those back.

mhscott:
Thanks. I know it's not a big deal in this case, but it looks like you manually re-did the changes from earlier today. (I can tell by the spacing). It would be better to use git commands. Imagine if you had to re-do multiple lines of code...

msalehi2004:
> Thanks. I know it's not a big deal in this case, but it looks like you manually re-did the changes from earlier today. (I can tell by the spacing). It would be better to use git commands. Imagine if you had to re-do multiple lines of code...

You are right :) - it was a minor change so I did it manually!

PR #883, opened 2022-07-04
Improve the local convergence for the UVCuniaxial model.

ahartloper:
There was a "sign(sigma - alpha)" term "missing" in the calculation of the denominator of the local Newton iterations for the UVCuniaxial material. This fix will slightly improve the local convergence rate with strain increments < 0.

### DIFF
--- a/SRC/material/uniaxial/UVCuniaxial.cpp
+++ b/SRC/material/uniaxial/UVCuniaxial.cpp
@@ -182,6 +182,7 @@ void UVCuniaxial::returnMapping(double strainIncrement) {
   double stressRadius = 0.;
   double phi = 0.;
   double ePEq = strainPEqConverged;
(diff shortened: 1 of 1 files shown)


ahartloper:
Also have the "OPS_GetIntInput()" and "OPS_GetDoubleInput()" changed in the past 2-3 years? In testing with a .tcl file, I found that
```
if (OPS_GetIntInput(...) != 0) {
\\ do stuff
}
```
would do stuff each time, but the inputs appeared to be correct and the analysis ran fine.

mhscott:
I think you want to use < 0 not != 0 for the OPS_Input functions

ahartloper:
> I think you want to use < 0 not != 0 for the OPS_Input functions

Ah great, thanks, I'll take a look at it!

PR #887, opened 2022-07-12
Update ASDEmbeddedNodeElement for UP

MassimoPetracca:
@fmckenna  @mhscott  @jaabell in this PR I added a new option in the ASDEMbeddedNodeElement to allow constraining also the P DOF for UP elements.
The documentation was updated accordingly in OpenSees/OpenSeesDocumentation#43


jaabell:
I went ahead and checked that it compiles OpenSeesMP (with Makefiles) and OpenSees and OpenSeesPy with CMAKE on Linux. Everything looks good.

PR #894, opened 2022-07-17
Creating spNode command and update sp command 

siwalan:
This pull request implement two things

spNode command, which will allow user to impose an sp constraint in the same way one put a load. e.g. spNode Node ndf1 ndf2 ... ndfN

Overload of the sp command that if it detects possible usage of sp command in the same way with spNode command, it will redirect to spNode command instead. Checked with several cases but I am not sure if it will cover all the possible edge cases.

I am interested in implementing the same thing in the Py interpreteter but I am stuck in finding the 'correct' way to get the domain NDF.

### DIFF
--- a/SRC/modelbuilder/tcl/TclModelBuilder.cpp
+++ b/SRC/modelbuilder/tcl/TclModelBuilder.cpp
@@ -270,6 +270,9 @@ TclCommand_addNodalMass(ClientData clientData, Tcl_Interp *interp, int argc,
 int
 TclCommand_addSP(ClientData clientData, Tcl_Interp *interp, int argc,   
 		      TCL_Char **argv);
+int
+TclCommand_addSPNode(ClientData clientData, Tcl_Interp* interp, int argc,
+	TCL_Char** argv);
 
 int
 TclCommand_addImposedMotionSP(ClientData clientData, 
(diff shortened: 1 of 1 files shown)


siwalan:
Script used to test (based on https://portwooddigital.com/2020/08/27/most-solvers-can-be-marplots/)

```
wipeAnalysis
model BasicBuilder -ndm 2 -ndf 3
set L 100
set E 29000
set A 20
set I 1400
node 1 0 0
node 2 $L 0
fix 1 1 1 1
fix 2 1 1 1
geomTransf Linear 1
element elasticBeamColumn 1 1 2 $A $E $I 1
timeSeries Constant 1

# pattern Plain 1 1  {
#     sp 2 2 1.0
# }

pattern Plain 1 1  {
    sp 2 0 1 0 -constant 0
}

constraints Transformation
system UmfPack
analysis Static
analyze 1
reactions
nodeReaction 1


```

mhscott:
The Python commands are in https://github.com/OpenSees/OpenSees/blob/master/SRC/interpreter/OpenSeesPatternCommands.cpp
You'll see other commands that get NDF

siwalan:
@mhscott  Thanks for the info. I added the Py command for spNode, but for the sp 'overloading'. I think it will be little bit tricky because after checking out the input (using OPS_GetXXXInput), I don't see any way to 'roll back' the pos again in Py.

PR #899, opened 2022-07-19
IGA: fix abs functions using fabs

zhuminjie:


jaabell:
Thanks! Was this giving a warning on VisualStudio?

PR #938, opened 2022-08-01
Build workflows supporting CI/CD

z-ichinohe:
Hello, OpenSees Moderators.

  Existing "build_cmake.yml" is not working, but Github Actions is a powerful tool to make our own build environments unnecessary. I made two workflows for MacOS and Linux. They are triggered manually or when updating each YML file. Product files such as OpenSees or opensees.so are provided as an artifact. Being uploaded to the branch or deployed to PyPi directly is preferable, but it has some difficulties for me. However, this build process automation will be very helpful for either users or developers. The workflow for Windows must also be necessary.
  Moreover /OTHER/metis/version was renamed to metis_version, in order not to have conflict with "version" included in a compiler.


siwalan:
If OpenSees plan to have some kind of CI/CD, it would be great if the version number compiled from Git has the Git version has written in the version. Based on some Google, this could be done by modifying the  cmake file

https://stackoverflow.com/questions/1704907/how-can-i-get-my-c-code-to-automatically-print-out-its-git-version-hash

z-ichinohe:
It seems good idea and must be necessary. It might be implemented by a modification on OPS_VERSION appearing in pythonMain.cpp and TclInterpreter.cpp, but I have no idea about where they get that.

jaabell:
I agree these things would be useful. Current work on conan/cmake is needed before any other automation can be added though. In the end, it would be up to @fmckenna to decide to go with something like this. You have my vote for what its worth.

PR #970, opened 2022-08-26
Improve output of MVLEM_3D, SFI_MVLEM_3D and FSAM material

MassimoPetracca:
@fmckenna @mhscott @kkolozvari 

This PR improves the output of **MVLEM_3D** and **SFI_MVLEM_3D** elements as well as the **FSAM** material.

- OPS_Stream tags and attributes are added where necessary to improve the metadata of the results
- theResult (Result*) object is returned only at the end: Returning it within each if-statement did not allow the following code (to close the previously open OPS_Stream tags) to be executed
- Fixed a small bug causing segmentation fault on Windows: In `return eleInfo.setVector(this->getShearDef());` the setVector was called (instead of the setDouble) with a double. It compiled fine probably because the double was implicitly converted to int, which was, in turn, converted to a Vector since the Vector class has a constructor taking int as an argument...
- In the FSAM I added a new output called panel_crack. It is a 2x2 tensor (a Vector(3) in Voigt notation) whose 2 eigenvalues are the crack-strains and the associated eigenvectors are the crack tangent vectors:
 `T = c1*outer_prod(T1,T1) + c2*outer_prod(T2,T2)`
It's useful for visualizing the crack pattern using any vector plot routine able to represent 2D tensors


kkolozvari:
@fmckenna @mhscott @MassimoPetracca looks good to me. Thanks.

PR #974, opened 2022-08-27
Fix "ascii art" drawing at the top of the files

lcpt:
This PR fixes two little mistakes in the drawing the author uses to explain the stress-strain diagram of this material.


jaabell:
Thanks!

PR #976, opened 2022-08-29
Adding Orthotropic Material Wrapper

MassimoPetracca:
@fmckenna @mhscott @jaabell 

With this PR I would like to add a new ND material wrapper called Orthotropic.
It is a simple wrapper that can convert any 3D isotropic material model into an orthotropic one, based on the _Transformed-Tensor-Method_:

- The input strain is mapped from the orthotropic (real) space to the isotropic (fictitious) space.
- Calculations are carried out in the isotropic space with any isotropic material model.
- The stress and tangent are mapped back to the orthotropic (real) space.

Documentation (with references) and a Tcl and Python example are provided in OpenSees/OpenSeesDocumentation#54.
The example performs multiple single-element-tests at different loading angles to build the yield surface in the plane-stress space for a J2Plasticity and its orthotropic wrapper:
Orthotropic_result


gaaraujo:
@MassimoPetracca  
Thanks for implementing this wrapper! I used it earlier this week and had a question about the implementation.

Is there currently a way to reproduce a tensor map $\mathbf{A}^{\sigma}$ with a block diagonal Voigt representation, rather than a purely diagonal one? In the [Oller2003](https://core.ac.uk/download/pdf/296535134.pdf) reference, the authors define $\mathbf{A}^{\sigma}$ as:

$$
\mathbf{A}^{\sigma} =
\begin{bmatrix}
B_{11}^{\sigma} & B_{12}^{\sigma} & B_{13}^{\sigma} & 0 & 0 & 0 \\
B_{21}^{\sigma} & B_{22}^{\sigma} & B_{23}^{\sigma} & 0 & 0 & 0 \\
B_{31}^{\sigma} & B_{32}^{\sigma} & B_{33}^{\sigma} & 0 & 0 & 0 \\
0 & 0 & 0 & B_{44}^{\sigma} & 0 & 0 \\
0 & 0 & 0 & 0 & B_{55}^{\sigma} & 0 \\
0 & 0 & 0 & 0 & 0 & B_{66}^{\sigma}
\end{bmatrix}^{-1}
$$

From the source code and docs, it looks like the current implementation uses a diagonal form:  
$\mathbf{A}^{\sigma} = \mathrm{Diag}(A_{11}^{\sigma}, \ldots, A_{66}^{\sigma})$.

I think a block-diagonal map should be straightforward to add but wanted to confirm whether there was a specific reason for choosing the diagonal map before trying to do a PR.

MassimoPetracca:
@gaaraujo 
Thanks for your comment and sorry for the late reply.

In the reference paper, in equation 36 you can see that A^-1 is the product of a diagonal matrix B, and an optional non-diagonal matrix alpha.

While B is sufficient to correctly map the points along the X-Y-Z axes of the Isotropic yield function to the points of the real Orthotropic one, it is not sufficient to map the whole function.

B is easy to build: it just requires the user to know the orthotropic strengths (easily available from tests).
However, alpha requires the user to know (and to input) the exact form of the yield crieterion. So it's more a theoretical aspect, but in practice no one will ever use it.

Furthermore, using only B, the wrapper is linear, while to obtain the values for alpha, the solutions becomes nonlinear

PR #978, opened 2022-08-31
Improvements to NDTest

jaabell:


jaabell:
In this PR, I did a wrong commit in 619e7e4 where I wanted to keep the remote CMakeLists.txt (the one last changed by fmk in commit 5d5ed1a950139568da58a2527b6bb67b7c75dac6). So I went ahead and copied the one on the main opensees repo and pushed that one. This keeps the changes by @fmckenna . 

Then I added the if statements to only look for MPI when its needed (otherwise I get compile errors when I have MPI but just want to compile regular opensees or openseespy). 

Then I fixed a small typo.

PR #980, opened 2022-09-02
Fixes to NDTest

jaabell:


jaabell:
Preventing evil segfaults

PR #984, opened 2022-09-06
Small fix to domain screen output with repeated tags. 

jaabell:


jaabell:
Hi Silvia. The legibility of the error report to the screen was improved. I don't see the problem.

silviamazzoni:
which is why someone shouldn't commit their own PR.

jaabell:
Thanks for the feedback Silvia. This is first I've heard about this rule of not fixing typos on classes. I've seen may other PRs fixing typos in the past. 

Regarding the merging of my own PRs. I used to leave my own PRs for them to get reviewed before getting merged into the OpenSees master branch. In the end, I was invited by @mhscott to merge my own PRs, as I have been given this permission by @fmckenna since the creation of the GitHub OpenSees project. I used to have SVN access and was also given immediate access when we moved on over to GitHub. Frank and Mike both merge their own PRs, as I do my own since this year when I added the new isogeometric analysis component and also for the H5DRM stuff and more. I try my best not to break other people's code or build, and no one has time to review my own work. 

What changed is that I recently tried my hand at being more active taking care of other pending issues and PRs to help out with the OpenSees project (it is a community in the end, and I feel I should contribute beyond my own interests as you and Mike do). In doing so I probably interfered with your work and Mike's, which I didn't mean to and apologize if it caused any trouble. All I want is to do is keep making OpenSees better, hopefully being more responsive to community requests by solving issues and merging PRs that meet the code standards. 

I hope, moving forward, we can communicate a bit better so we know who is working on what. That way we can all collaborate without stepping on each other's work. A mechanism for this is to use GitHub's "assign" developers to issues and PRs, which is standard on most big projects. 

What do you think?

Thanks again.  

Sincerely, 
José

PR #989, opened 2022-09-09
Changing ARPACK parameters so that eigen will work in a few more cases

mhscott:


jaabell:
Interesting.  I was just using the eigensolver for a case I knew had rigid body modes, and it failed. So I ended up using your excellent `GimmeMCK` integrator to get the stiffness out and using `scipy.eigh` to do the eigenvalue analysis. 

Context: I'm teaching my students how to "debug" a truss which is accidentally a mechanism by using eigen analysis on the system stiffness.

Do you think the eigensolver is now capable of finding rigid body modes?

mhscott:
No. But it sounds like you're talking about eigenvalues of K only, which is best done with a different solver:
https://portwooddigital.com/2020/11/13/ordinary-eigenvalues/

jaabell:
Indeed! I had no idea about that. Thanks!

fmckenna:
fmk - try and make arpack fail if solver fails instead of givng numbers that are useless

PR #995, opened 2022-09-15
Check for MPI only for OpenSeesMP and OpenSeesSP targets

jaabell:
On Linux, when using CMake, if I have the MPI compilers it will set `MPI_FOUND` to `True` which in turn enables looking for Mumps even if my compilation target is `OpenSees` or `OpenSeesPy`. 

So this PR restricts looking for MPI only when the target is `OpenSeesMP` or `OpenSeesSP`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,8 +260,11 @@ endif()
 
 find_package (Python COMPONENTS Interpreter Development)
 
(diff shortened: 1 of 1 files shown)


fmckenna:
have updated instructions in online doc showing how to install mumps and listing all packages required to build the parallel ones.

PR #999, opened 2022-09-27
New output APIs

zhuminjie:
Add following output APIs

1. OPS_SetIntListsOutput - return a list of list of int
2. OPS_SetIntDictOutput - return a dict of int
3. OPS_SetIntDictListOutput - return a dict of list of int
4. OPS_SetDoubleListsOutput - return a list of list of double
2. OPS_SetDoubleDictOutput - return a dict of double
3. OPS_SetIntDoubleListOutput - return a dict of list of double
4. OPS_SetStringList - return a list of string
5. OPS_SetStringLists - return a list of list of string
6. OPS_SetStringDict - return a dict of string
7. OPS_SetStringDistList - return a dict of list of string


fmckenna:
We need the updates for the Tcl interpreter as well before I can accept this.

fmckenna:
we need implementations of these new API function for when people start using these new APIs .. think code needs to go into: elementAPI_TCL.cpp .. tcl has a dict.

zhuminjie:
Okay, I'll try to implement that

zhuminjie:
Add the APIs to elementAPI_TCL.cpp for Tcl.

PR #1002, opened 2022-09-30
Adding Concrete02 to FSAM nDmaterial

kkolozvari:
This PR allows FSAM to use Concrete02 (so far it worked only with ConcreteCM). This will improve numerical stability and efficiency of the FSAM material and SFI-MVLEM and SFI-MVLEM-3D models.

Please let me know if you have any questions. Thank you.


mhscott:
@kkolozvari Please add `mon` to sendSelf and recvSelf

kkolozvari:
@mhscott thanks for catching that. I just made a fix. Please let me know if there is anything else. Thank you.

mhscott:
@kkolozvari If your intent with `mon` is to make Concrete02 or ConcreteCM monotonic, then FSAM::commitState() can simply *not* call commtState() on theMaterial[2] and theMaterial[3]. This way you don't have to add `mon` variables to Concrete04 and Concrete23, etc.

kkolozvari:
@mhscott that could probably work too, but I will have to test it. I am not sure if other concrete materials have clearly defined envelope function like ConcreteCM and Concrete02 and I am not sure what will happen if I don't commit materials (never tried this before). At this time I dont plan to add other concrete materials to FSAM and this implementation should work OK. Is _mon_ parameter not acceptable?

mhscott:
@kkolozvari It seems like not calling commitState is the way to go even though you don't plan to add other materials right now. I'm sure you will later as people continue to use FSAM for wall models. All the concrete materials have a backbone function, so that's not a concern.

kkolozvari:
@mhscott sure thing. I already implemented it, but I need to do some more testing. it would be definitely a more elegant solution and the code would be cleaner too. I should be able to submit another PR in a couple of weeks. Thanks for your suggestions.

fmckenna:
how is the update coming?

kkolozvari:
@fmckenna Its on my radar, but I got sidetracked with some other urgent items. I will work on it in coming weeks and hopefully submit a PR soon. I just need to do more testing...

kkolozvari:
I have done some testing with the formulation that doesn't commit monotonic materials, but results are worse (comparing to test data) and hysteretic shape is a bit strange for some cycles. I think it is better to stick to the original formulation. If you agree, I will submit the PR without the latest commit where I removed monotonic materials from commitState. Please let me know what you think. Thanks.

kkolozvari:
@mhscott @fmckenna hello, can you please let me know what is the status of this? Thank you.

mhscott:
@kkolozvari Thanks for the reminder. The FSAM and ConcreteCM/02 relationship is pretty fragile, e.g., one could change the order of return values from ConcreteCM/02::getInputParameters() and totally break the FSAM code. This setting of parameters should be manageable through input to the FSAM constructor (after all, you're only getting a few basic parameters out of ConcreteCM/02), not through implicit assumptions about what type of material is used by FSAM. What will you do when you or someone else wants to make other concrete materials work with FSAM?

kkolozvari:
@mhscott FSAM is really just a shell that uses uniaxial materials, so it is dependent on them and their input. In case you described some changes to FSAM would have to be done, but they would be minimal. I agree that implementation it is not perfect (nothing is), but having FSAM working with Concrete02 improves its numerical stability and efficiency a lot. I have been using this code for years and published many papers based on it, and I just decided to release it to the public so others can do the same. Can you please suggest a resolution of this item? Thank you.

mhscott:
@kkolozvari You should pass the necessary parameters into the FSAM constructor.
Yes, nothing is perfect. But publishing many journal papers on something is not a proxy for models that are flexible and extensible.

kkolozvari:
@mhscott  In formulation of FSAM there are total of 8 copies uniaxial concrete materials that work through 3 stages of RC panel behavior. The model needs to have these copies of uniaxial materials and they are not only used to obtain input parameters. I don't understand exactly what is the issue and how to solve it. Can you please be more specific? Would be happy to jump on the call... Thanks

paysheen:
@kkolozvari 
Is it possible now to use concrete02 in FSAM?

kkolozvari:
@fmckenna and @mhscott, would you please reconsider accepting this PR? I tried removing _mon_ variable, but the model simply doesnt work the same way. There is some interest from the engineering community in having this model available publicly. Please let me know. Thanks,

mhscott:
Consider making the cracking strain an input to FSAM instead of attempting to read from materials internally

kkolozvari:
@mhscott @fmckenna With ConcreteCM, the model uses cyclic cracking strain, with Concrete02, the model uses stress-based criteria because cyclic cracking strain is not available in the model formulation. Please note that there are other nDmaterial models in OpenSees that are "tied" to a specific uniaxial concrete material. FSAM worked only with ConcreteCM for 10 years now, so I am not clear why there need to be changes in the input after all these years. I am just trying to add Cocnrete02 to increase its stability and numerical efficiency, which was done in the exactly the same manner as for ConcreteCM. Thanks

mhscott:
You are asking for input parameters via internal function calls. Why not let the user pass the necessary ConcreteXYZ input parameters into FSAM so that FSAM can be used with any uniaxial material and we will head off the need for additional incremental PRs that do nothing more than add getInputParameters to Concrete03, Concrete04, etc?

And just because something has been this way for years doesn't make it OK to keep doing the same thing. Let's improve OpenSees by doing things better.

silviamazzoni:
Kristian, it's a long-term vision issue.
MVLEM would have to be updated every time there is a new concrete material. and there are many. I always recommend taking the extra development time to ensure robustness in the long term.

fmckenna:
Kristian,

I looked at the code .. What you submitted is not going to work .. You do not use the getInputParameters you added to setResponse in Concrete02 and use a getSOMETHING related to cyclic strain in FSAM that is not in the Concrete02 material submitted.

kkolozvari:
Hi Frank, Thanks for your response. I dont understand the second part of your comment related to "getSOMETHING". Cyclic strain is not available in Concrete02 and FSAM is not asking for it. For Concrete02, the criterion for the 2nd crack is stress-based, not strain-based, exactly for this reason. 

Can you suggest a solution that does not involve doubling the input parameter for both uniaxial materials and FSAM? I could rewrite the code using getParameter or other similar existing functions in OpenSees so I don't disrupt the Concrete02 code too much.

Please let me know. Thanks

mhscott:
Doubling input parameters is a relatively minor issue and will not require _any_ modification to Concrete02 or any other stress-based model that comes along.

```tcl
ft = 4
Ets = 4000
et = ft/Ets

uniaxialMaterial Concrete02 1 ..... ft Etc .... blah
nDMaterial FSM 2 1 ...... -et et
```

fmckenna:
sorry .. I had to expand the github source code to show all the lines of the file .. if you just update your repo to latest so that I know it works and resubmit .. when time in future you can look into possibility of making it avail to all concrete materials by adding a setResponse method that gets what you need from all materials so any concrete material will work with FSAM.

kkolozvari:
@fmckenna will do. Thank you.

PR #1005, opened 2022-10-01
Error handling with responseSpectrum command

mhscott:
Sudden termination (dead kernels) via exit(-1) if modalProperties has not been called prior to responseSpectrum.
Modified the Domain::getModalProperties() method to return 0 (success) or -1 (error).
Tested on simple OpenSeesPy example and no weird side effects.
We'll wait for @MassimoPetracca's approval to merge.


MassimoPetracca:
@mhscott It sounds good to me. There should be no side effects.

PR #1024, opened 2022-10-12
fix static-analysis stiffness and memory leak with damping

yulee:
Set unity stiffness modifier by damping for static analysis; fix memory leak


mhscott:
@yulee Aren't you also producing a memory leak when calling setDamping, e.g., with ASDShellQ4 when m_damping[i] is already set to a damping model? Probably similar issues in the other elements you modified
image

yulee:
Great catch! I went ahead and added the deletion of the existing damping instance.

zhuminjie:
It's also good to return error code to flag errors, like return -1 here, instead of exit, which will crash without giving any warning

mhscott:
You should return negative values for error conditions, not positive, .e.g, -1 or -2 instead of 1 or 2

PR #1027, opened 2022-10-14
add some py models in Tcl

zhuminjie:


BijanSeif:
Hi @zhuminjie and also @mhscott 
I curious to know that why we do not automate adding commands instead of writing what you committed in this request. I mean instead of writing _extern void* ..._ and _else if strcmp ..._ because all of these lines are repetitive, we read the files available in the directory and then get the names of the objects and then create these commands automatically? By this way there is no need to add fill this file when some new models added by developers. is there any special reason not to do it?

zhuminjie:
I don't think there is an easy way to do what you said. Can you show an example how to do it?

BijanSeif:
Okay. Because it should be shown on a code platform and it is hard to show it by texting, Friday I will meet Prof. @mhscott, and I will show him some samples that I implemented before on python.

PR #1028, opened 2022-10-15
221015 New HystereticSM material

silviamazzoni:
Here is my first official pull request!
I have extended the Hysteretic-material envelope to have a maximum of 7 points, compared to the 3 in the original.
I have also added some recorder options that make it very practical for hinge modeling.


silviamazzoni:
I don't think the failed checks are due to my pull.....

MassimoPetracca:
@silviamazzoni It does not find your header file:

image

Since it is using **CMake** to compile OpenSees, you just need to add your H and CPP files in the **SRC/material/uniaxial/CMakeLists.txt** file (as it is done for the original HystereticMaterial):
image


You should also support old make files, so I would also add them here:
image

PR #1039, opened 2022-10-23
Adding command to return list of pattern tags

mhscott:


BijanSeif:
@mhscott Thats useful, But what about the pattern types (Nodal or linear) and values? That would be great if a command be there that ask the pattern tag and return the nodes or elements loads values (I don't know if there is)

mhscott:
Yes, those commands are here, added by @sewkokot 
https://github.com/OpenSees/OpenSees/pull/575

PR #1041, opened 2022-10-25
PFEMContact3D

zhuminjie:
1. PFEMContact3D - a 3D contact element for PFEM background mesh
2. Flume - a mesh of wave flume 
3. BeamBrick - a mesh of beam that creates a brick/rectangle


fmckenna:
can you add the documentation to OpenSeessDocumentation .. probably jus copy the rst file from openseespy doc

PR #1043, opened 2022-10-27
Flush recorders when analyze is done

zhuminjie:


fmckenna:
@mhscott @zhuminjie make as option to analyze command.

zhuminjie:
The '-noFlush' option is added

PR #1054, opened 2022-11-10
BugFix in PDMY03 material

MassimoPetracca:
@mhscott @fmckenna @jaabell 

While trying to find the reasons for a strange behavior in PDMY03 with @jinchilu, we found and fixed few bugs in the computation of the initial stress, and in the send/recv methods


jaabell:
Hi @MassimoPetracca. Can you maybe provide an example? I'd be happy to check the code and do the merge.

MassimoPetracca:
@jaabell sure!

Let me explain the issue.
the **stress0** variable (representing the stress stage at the end of the gravity analysis) was computed as a static variable, therefore it was evaluated only once per process.
`static Vector stress0 = currentStress.t2Vector();`
Doing so, it was actually representing the stress0 of the first element in each process and not the stress0 for each element.

In sequential analysis, this bug was hardly visible. Instead, in parallel analysis it becomes evident:
image
The results are different from the 1-process model and the 8-process mode (in this example).
And you can clearly see stress jumps coinciding with the partition of the mesh.

Here you can find both the SCD (STKO pre-processor file) and the TCL files for a model to be run with OpenSeesMP with 8 processes
[model_pdmy03_8p_new_2.zip](https://github.com/OpenSees/OpenSees/files/9987985/model_pdmy03_8p_new_2.zip)

fmckenna:
@jaabell jose .. how is the testing coming?

PR #1058, opened 2022-11-15
Another mesh type used for 3D PFEM 

zhuminjie:


fmckenna:
@zhuminjie you need to fix the build issue.

zhuminjie:
Yeah, the build error is like the following, there are many of them but similar

/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libpython3.8.a(getcompiler.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC

Do you have any idea what they are?

mhscott:
@fmckenna Is `-fPIC` enabled in the CMake settings?

zhuminjie:
There is nothing special here. How about I close this one and resubmit again.

PR #1059, opened 2022-11-16
Resubmit mesh type for PFEM 3D

zhuminjie:


zhuminjie:
failed with following error

zlib/1.2.12: Downloaded recipe revision 0
WARN: Remotes registry file missing, creating default one in /home/runner/.conan/remotes.json
ERROR: Conflict in tcl/8.6.10:
    'tcl/8.6.10' requires 'zlib/1.2.12' while 'hdf5/1.12.0' requires 'zlib/1.2.13'.
    To fix this conflict you need to override the package 'zlib' in your root package.
Error: Process completed with exit code 1.

fmckenna:
that is on me .. as the other one might be!

PR #1063, opened 2022-11-20
Correction on HystereticPoly material

jotru:
@salsessa setTrialStrain function doesn't seem to have an if-else statement as shown in the paper "An accurate and computationally efficient uniaxial phenomenological model for steel and fiber reinforced elastomeric bearings". Corrections were made according to the formula stated in the paper.


salsessa:
The paper's algorithm is obsolete. 
Actually, if the condition of the (new) line 180 is not fulfilled, the difference between expressions of lines 182-185 and lines 190-191 is smaller than 1.0e-20 (Since equation of line 190 is the asymptote of Tstress in line 182).
Hence, the code is a lot faster if the "if then else" condition is avoided.

mhscott:
@salsessa If the paper is obsolete and does not accurately portray the implementation, update the documentation accordingly.

mhscott:
@jotru If you can show cases where the if/else statement makes a noticeable different, please post plots here.

salsessa:
I just made a pull request with the documentation update:
https://github.com/OpenSees/OpenSeesDocumentation/pull/68

jotru:
@salsessa @mhscott  
I have attached the screenshot of the two models. As shown, there is a significant difference between the corrected code and the current implementation in OpenSees. The parameters used are also shown in the screenshot. 

image
image

salsessa:
@jotru @mhscott
Dear Jonathan,
Parameter "tol" is definitively incorrect. It is an asymptotic tolerance, determining when the transition curve reaches the boundaries of the hysteresis loop. We used 10^-20 although 10^-8 is sufficient. However, tol=5 it’s an excessively huge tolerance and is definitively wrong.
If you assume tol=10^-8, you get these responses:
DeltaE-8


And their numerical difference is less than 10^-32.
The original formulation was slightly modified in these references:
https://link.springer.com/article/10.1007/s00161-021-01044-w
https://link.springer.com/chapter/10.1007/978-3-031-04548-6_27
that I added to the documentation page https://github.com/OpenSees/OpenSeesDocumentation/pull/68 . This was in order to avoid the “if..else” statement that slows down the code.
Moreover, your parameters look a bit strange to me. B_1_1 and b_2_1 seem to be very small as well as alpha. Have you tried to identify them by the freeware available here: http://bit.ly/35F5x7Q  ? However, if you need any support to calibrate the parameters, you can write me a private message and I’ll be happy to help you.

jotru:
@salsessa
 Dear Salvatore Sessa, 
I have read the paper and I have seen the tolerance used in the papers so my first trial was to fit the curve without the use of the optional parameter tolerance. But I couldn't get any parameter that will fit my curve. That's why I resorted to using the optional tolerance parameter.  I have also tried the freeware which I assume is also linked in the OpenSeespy documentation page but it didn't work for me so I had to use my own optimization algorithm to find the parameters. 

 I assumed the model is a phenomenological model which we can freely set to fit our curve. Are there any restrictions on the parameter values we can use? Note that the units I used in my model (KN, mm) are different than the curves fitted in the paper(N m), and the material strength is also significantly different so how will these restrictions work with different units?

I will be happy if you can assist me with calibration to find a better fit but I don't know how to contact you personally. I tried to find a way to do it on GitHub but I couldn't seem to find a way :). if you kindly text me your email address, I will contact you Asap.

salsessa:
@jotru write me an email at salvatore.sessa2[at]unina.it and send me a table with your experimental data, I'll try to support you.

fmckenna:
@jotru @salsessa just update this thread when you reach a conclusion

salsessa:
@jotru @mhscott @fmckenna 
Dear all,
We tested the code and it is actually consistent with the papers reported in the updated documentation https://github.com/OpenSees/OpenSeesDocumentation/pull/68.
Parameter "tol" is a tolerance, it must be very small and it must be definitely smaller than 1.
The identification process fails in determining the parameters of the experimental loop provided by @jotru because such an hysteresis presents pinching, and the model cannot address it. A different pinching model is currently under investigation but it has not been implemented yet.

If "tol" is set smaller than 1.0E-6, as it is reported in the references, the currently implemented AsymmetricPoly (that is consistent with this paper: https://doi.org/10.1007/s00161-021-01044-w ) provides exactly the same results of the code committed by @jotru (that is consistent with this older and obsolete reference https://doi.org/10.1016/j.compstruct.2018.12.017 ).

Actually, the only difference between the two formulation is that the older one requires an "if..then" instruction that significantly slows the code. For this reason, whe choosed to implement the new version.

jotru:
@salsessa 
Dear Professor Salvatore, 
Thank you for the explanation and I will investigate further if a better fit is possible.

PR #1064, opened 2022-11-20
New materials are added, Shear01 and Axial01

erkanbicici:
Two uniaxial materials are added to calculate lateral shear and axial displacements of a reinforced concrete column.

### DIFF
--- a/SRC/material/uniaxial/Axial01.cpp
+++ b/SRC/material/uniaxial/Axial01.cpp
@@ -1,0 +1,426 @@
+/* ****************************************************************** **
+**    OpenSees - Open System for Earthquake Engineering Simulation    **
+**          Pacific Earthquake Engineering Research Center            **
+**                                                                    **
+**                                                                    **
+** (C) Copyright 1999, The Regents of the University of California    **
+** All Rights Reserved.                                               **
+**                                                                    **
+** Commercial use of this program without express permission of the   **
+** University of California, Berkeley, is strictly prohibited.  See   **
+** file 'COPYRIGHT'  in main directory for information on usage and   **
+** redistribution,  and for a DISCLAIMER OF ALL WARRANTIES.           **
+**                                                                    **
+** Developed by:                                                      **
+**   Frank McKenna (fmckenna@ce.berkeley.edu)                         **
+**   Gregory L. Fenves (fenves@ce.berkeley.edu)                       **
+**   Filip C. Filippou (filippou@ce.berkeley.edu)                     **
+**                                                                    **
+** ****************************************************************** */
+                                                                        
+//
+// $Date: 2016/05/29 12:19
+// 
+// Written: Erkan Bicici 
+//
+// Description: Axial deflection model with respect ot lateral deflection 
+//				02/11/2018 Applicaiton of New features. 
+//				Eliminate zero stiffness.
+
+#include <elementAPI.h>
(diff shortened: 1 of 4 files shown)


fmckenna:
need some documentation with examples added to the openseesdocumenation (https://github.com/OpenSees/OpenSeesDocumentation)
repo before we will accept. is this material only for zero length elements? do you envision it going into fiber sections??

mhscott:
@erkanbicici It is unusual that these uniaxialMaterials need to keep node pointers. You may be using the wrong abstraction. Perhaps a zeroLength or twoNodeLink element would be better? Are you attempting similar functionality to the LimitCurve abstraction? Documentation and examples with explanations would be helpful.

erkanbicici:
[Usage Description for Shear01 and Axial01.pdf](https://github.com/OpenSees/OpenSees/files/10065384/Usage.Description.for.Shear01.and.Axial01.pdf)

erkanbicici:
> need some documentation with examples added to the openseesdocumenation (https://github.com/OpenSees/OpenSeesDocumentation) repo before we will accept. is this material only for zero length elements? do you envision it going into fiber sections??

Prof. McKenna,
The materials are developed to use in zero-length-elements. I have never used them in fiber section. 
I shared a file that describes the usage of the models. 

Do you think I should share an actual model (.tcl file) as an example?

fmckenna:
@erkanbicici still awaiting pull request to OpenSeesDocumentation repo .. and yes a tcl file would be good as an example .. also need a tcl file in the EXAMPLES/runVerification directory that we can set to run automatically so we can check that the builds work.

erkanbicici:
> @erkanbicici still awaiting pull request to OpenSeesDocumentation repo .. and yes a tcl file would be good as an example .. also need a tcl file in the EXAMPLES/runVerification directory that we can set to run automatically so we can check that the builds work.

I have just pull requested in OpenSeesDocumentation. However, I am not sure I did it accurately. To be honest, I am not familiar with the usage of .rst and .html so I uploaded .pdf file. If there is something I need to change, please let me know.
Thank you.

mhscott:
@erkanbicici I'm sorry, the more I look at these classes, I am convinced they *should not* be uniaxial materials. They should be elements as they are obtaining displacements from nodes. They need not be zero length or two node link elements, they can be their own kind of element, but they just do not make sense as a uniaxial material.

mhscott:
Maybe these should be subclasses of LimitCurve.
UniaxialMaterial is not the right abstraction.

erkanbicici:
@mhscott I get your point. I think you are right. But what should I do? 
Just changing the folder or change the code completely?

mhscott:
@erkanbicici I think you should use a different abstraction. Take a look at the LimitCurve implementations.

erkanbicici:
> @erkanbicici I think you should use a different abstraction. Take a look at the LimitCurve implementations.
Sorry about late response. 
@mhscott I have been considering to turn the Axial01 to LimitCurve implementations. However, the proposed model does not totally works as LimitCurves. The model gives stress for a given strain with respect to shear displacement of column (location of shear spring in defined in the model). That means, this model is not defines a curve.
Maybe we can consider it as uniaxial material.

mhscott:
@erkanbicici Sorry, uniaxial materials should not be referring to nodes. This is the job of an element or some other abstraction.

Issue #1068, opened 2022-11-24
cNodeTag argument Not working for getRetainedNodes 


BijanSeif:
Hi,

In the [OpenSeesOutputCommands.cpp,](https://github.com/OpenSees/OpenSees/blob/master/SRC/interpreter/OpenSeesOutputCommands.cpp) command getRetainedNodes or getConstrainedNodes get one argument. Here is the part that it checks it.

```
    if (OPS_GetNumRemainingInputArgs() > 2) {
	  if (OPS_GetIntInput(&numdata, &cNodeTag) < 0) {
		opserr << "WARNING getRetainedNodes <cNodeTag?> - could not read cNodeTag\n";
		return -1;
	  }
	all = 0;
    }
```

It seems that it asks for the Constrained node tag, and it returns the retained node that cNodeTag is retained to. In the follwoing python code it does not work:

```
ops.wipe()
ops.model('basic','-ndm',3)

i=0
ops.node(1,0,0,0)
ops.node(2,1,0,0)
ops.node(3,2,0,0)
ops.node(4,3,0,0)
ops.node(5,4,0,0)
ops.node(6,5,0,0)
ops.node(7,4,0,0)
ops.node(8,5,0,0)
ops.node(9,4,0,0)
ops.node(10,5,0,0)


ops.equalDOF(8,9,1)

ops.rigidLink('bar',1,2)
ops.rigidLink('bar',1,4)
ops.rigidLink('bar',3,5)
ops.rigidLink('bar',3,6)



print('Constrained=',ops.getConstrainedNodes(1))
print('Retained=',ops.getRetainedNodes(2))
```

and its results is:

Constrained= [2, 4, 5, 6, 9]
Retained= [1, 3, 8]

As you see it returns all Constrained or Retained Tags instead of those that node2 has constrained to or node1 has retained to!

sewkokot:
@BijanSeif, Thank you for reporting this bug. I will fix it and send a PR.

BijanSeif:
Thanks, dear Dr. @sewkokot

PR #1069, opened 2022-11-24
Fix getRetainedNodes and getConstrainedNodes commands

sewkokot:
This PR is to fix https://github.com/OpenSees/OpenSees/issues/1068
@BijanSeif, I have tested it with your MWE, but more testing would be welcome.

### DIFF
--- a/SRC/interpreter/OpenSeesOutputCommands.cpp
+++ b/SRC/interpreter/OpenSeesOutputCommands.cpp
@@ -1669,7 +1669,7 @@ int OPS_getConstrainedNodes()
     int rNodeTag;
     int numdata = 1;
 
-    if (OPS_GetNumRemainingInputArgs() > 2) {
+    if (OPS_GetNumRemainingInputArgs() > 0) {
 	  if (OPS_GetIntInput(&numdata, &rNodeTag) < 0) {
 		opserr << "WARNING getConstrainedNodes <rNodeTag?> - could not read rNodeTag\n";
 		return -1;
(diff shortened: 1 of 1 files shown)


BijanSeif:
Again Thanks, Dr. @sewkokot.
As I understood, for OpenSees the equalDOF and RigidDiaphragm and RigidLink are same and define constrained and retained nodes (Correct me if I'm wrong).
So OpenSees do not record the type of command that nodes are retained or constrained to each other. I mean that we can't understand that nodes are constrained and retained to each other by Rigidlink or by equalDOF or by RigidDiaphragm while it is important to distinguish Rigidlink case from the other cases. (For plotting purposes, it is necessary to show the rigidlinks but for other cases is not)
let me know your idea and also guide me if there is a way to distinguish it.

Thanks.

sewkokot:
@BijanSeif, You can read more on this subject in https://github.com/OpenSees/OpenSees/issues/729
and in particular the ideas by @ambaker1 in https://github.com/OpenSees/OpenSees/issues/729#issuecomment-992756226

The MP constraint matrix can be displayed, for instance, using `ops.printModel()`.

However those ideas have not been implemented yet.

BijanSeif:
Dear @sewkokot 
I got my answers from your complete conversation. I would like to know that what happened about "getMPType"? "getMPType" is exactly what I am searching for? Could you add it to OpenSees? 
you said "However those ideas have not been implemented yet" 
is there any limitation for this purpose? or just finding free time!

sewkokot:
@BijanSeif, Adding the `getMPType` command would require more modifications in OpenSees source code than providing the `getConstraintMatrix` output command which then could be used to identify the MP type. I'm implementing the `getConstraintMatrix` now, and will see if it's enough for the purpose of opsvis. 

Also, there are limitations in Matplotlib to develop custom markers for OpenSees elements (equalDOF, rigidLink and RigidDiaphragm). And I simply need more free time to work on these features :)

BijanSeif:
Again, thanks dear Dr. @sewkokot for the answer.

BijanSeif:
Hi 
The above commands are working correctly in OpenSeesPy. I know that users should wait for OpenSees.exe release till use them in TCL 

But

getRetainedNodes AND getConstrainedNodes commands seems are not available for TCL commands!

PR #1077, opened 2022-11-30
New Ten-node tetrahedron element

jaabell:
This PR implements a new ten-node tetrahedron element into OpenSees. The element node numbering is
image
and has 4 Gauss-points inside forming a smaller 4-node tetrahedron inside with numbering following the first four nodes of the main-element node numbering.


jaabell:
Checked to compile under windows and linux. CMakeFile was made equal to current one in main opensees branch on commit e356ec2feae3bdfb2330d624cafa5022ec810165

PR #1079, opened 2022-12-01
Fix calculation of `tamax` in `sdfResponse`

otaithleigh:
The wrong iterating value was being used to calculate `tamax` -- the internal Newton `iter` instead of the index of the applied force `i`.


mhscott:
@otaithleigh Thanks!

PR #1165, opened 2023-02-17
Update how dummy parameter tag is defined in updateMaterialStage command

mhscott:


mhscott:
Prior to this PR, the following code will cause an error

```python
ops.parameter(2)
ops.updateMaterialStage('-material',1,'-stage',0)
```

PR #1172, opened 2023-03-11
fix a bug and improve connection setup in UDP socket

aschellenberg74:


fmckenna:
thanks. we need to get you to switch to cmake.

PR #1193, opened 2023-05-04
Adding factoronce to the Modified Newton and enhancing the embeddemBeaminterface elements implementation

amnp95:
This pull reuest include adding the factorOnce option or modified newton algorithm and enhancing the Implementation of embeddedInterfaceelements by implementing recvself() and sendSelf() for parallel processing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,5 +90,6 @@ Concrete.h
 forumat.for
 build
 bin
+CMakeListsmy.txt
 __pycache__
 *.dat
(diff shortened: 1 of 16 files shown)


mhscott:
Please move the #defines, e.g., `#define SL_NUM_NODE 4`, from the cpp files to private enum statements in the h files. You are redefining pre-processor directives, which some compilers do not like.
e.g.
```cpp
private:
   enum {SL_NUM_NODE=4};
```

amnp95:
Hi I just fixed the macros

zhuminjie:
There is an API, OPS_GetDomain, https://github.com/OpenSees/OpenSees/blob/2edbabe6f99f24a53fc370f81315704d07971de3/SRC/api/elementAPI.h#L176

that can be used to get a pointer to the domain. Should work in Tcl and Python

mhscott:
Thanks, @zhuminjie!

@amnp95 You should do something like this in your code:
```cpp
Domain* theDomain = OPS_GetDomain();
```

PR #1233, opened 2023-07-03
New material model and element

RasoolGhorbaniGitHub:
A new material model (GMG_CMAC2D) and element (GMG_CMAZ2D) were added. The material model and element are intended for use within a two-dimensional lumped-plasticity framework and aimed at capturing the lateral cyclic behavior of non-retrofitted and retrofitted reinforced concrete columns subjected to seismic motions, up to complete loss of lateral strength.

For more information please refer to my PhD dissertation - Chapter 3; 
Ghorbani, R. (2022). Computational Framework for Decision-Oriented Reinforced Concrete Column Simulation Capabilities. Ph.D., The University of Texas at San Antonio.

- Documentations for both the material model and element were added.

Thank you,
Rasool Ghorbani


mhscott:
What is the difference between this zero length element and the existing zero length element in OpenSees?

RasoolGhorbaniGitHub:
Hello Prof. Scott,
The new material model (CMAC2D) is capable of accounting for coupling behavior between all the Axial/Shear/Flexure (ASF) DOFs. For that, I need a zero-length (ZL) element that can send a vector of strains for all the ASF DOFs to the CMAC2D material model at every increment without needing any direction as an input (because technically it considers all DOFs). I modified the existing element such that it asks for **one material tag** (associated CMAC2D material tag), **without asking for directions**, constructs the corresponding vector of strains, and sends it back to the material model. The existing ZL element requires three material tags, each with a corresponding direction (which I couldn't use with the CMAC2D material model) to construct the vector of strains for ASF DOFs.

mhscott:
This sounds like something better implemented as a _Section_ object to be used in conjunction with the _zeroLengthSection_ element instead of creating a new element that is an incremental change from the current _zeroLength_ element.

RasoolGhorbaniGitHub:
Dr. Scott,
I think I need to provide more background on our new material model and element before I start implementing roughly 10000 lines of code in a new class.
Our material model gathers all the axial/shear/flexure DOFs in one place (we don't have separate uniaxial materials). The main reason is that we have defined a complex coupling behavior between all the actions such that each DOF should share information with the other DOFs during the analysis. Even the sequence of which DOF should be executed first dynamically changes during the analysis depending on which DOF needs what information from other DOFs.
I understand that by using ZeroLengthSection element we have the option to aggregate separate Sections and group them into one, but still, we would have faced the same issues to accomplish our goals. First, we cannot dynamically change the sequence of which material should be executed first using ZeroLengthSectoin. Second, we don't have separate material models for each DOF, we have only one material model that aggregates all. By using ZeroLengthSection element we are limited to defining Uniaxial Section in which we need to specify the direction (the string input).

silviamazzoni:
the ZeroLengthSection will take a fiber section, so it should work for an ND section.

mhscott:
@RasoolGhorbaniGitHub I'm curious how you estimated 10,000 lines of code for this endeavor. The _ZeroLength_ element has about 2,000 lines of code, _ZeroLengthSection_ about 1,000.

Anyway, I'm not talking about using _SectionAggregator_, I'm talking about writing a *new* Section model that does exactly what your _ZeroLengthGMG_ element does, but without all the unnecessary code duplication that comes with essentially duplicating the _ZeroLength_ element. If you would like to discuss the details further, we can set up a Zoom and I can show you what I'm talking about, which is difficult to do on this GitHub issue.

RasoolGhorbaniGitHub:
@mhscott 10000 is the number of lines for the material model, not the element. I thought by changing the ZeroLengthGMG to ZeroLengthSection, I also need to implement our material model (which is developed under Uniaxial Material class) into a "Uniaxial Section" class (which actually I suspect this would be the case as we have coupled our new material model to the new element).
That would be fantastic Dr. Sott, really appreciate your time in advance.

silviamazzoni:
Yes, I think it should be a section behavior because that's what it is both in terms of OpenSees as well as the mechanical model

mhscott:
` ELE_TAG_GMG_CMAZ2D` is not defined in classTags.h nor anywhere else. Are there missing pieces from this PR?

RasoolGhorbaniGitHub:
@mhscott I just rechecked my source code, I defined it in front of line 852 as **#define ELE_TAG_GMG_CMAZ2D           99992**. Not sure why I cannot find it here (maybe forgot to save that file before pushing the source code to the upstream) 
Please let me know if I push my source code here again. 
Thanks

mhscott:
@RasoolGhorbaniGitHub Yeah, probably forgot to save. Go ahead and push the updated classTags.h to this branch.

RasoolGhorbaniGitHub:
Got it. Will do, thanks.

RasoolGhorbaniGitHub:
@mhscott I just added the **ELE_TAG_GMG_CMAZ2D** to classTags.h.
Thanks again.

mhscott:
@RasoolGhorbaniGitHub After talking more with @fmckenna and @silviamazzoni this morning, this PR cannot be accepted because you modified the _UniaxialMaterial_ interface with a method that basically does _SectionForceDeformation_ behavior.

As discussed in our meeting last week, the P,M,V material model needs to be implemented as a subclass of _SectionForceDeformation_ and you should use the current _ZeroLengthSection_ element (which should not require any modification). Most of the code from your UniaxialMaterial implementation can be copied directly to a new section implementation. Your model is fundamentally the same as _Bidirectional_ and _Elliptical_ section models which couple section forces directly (no fibers).

Let me know if you would like to meet again to go over the _SectionForceDeformation_ interface.

PR #1234, opened 2023-07-03
Putting tension and compression hardening laws directly in OPS function

mhscott:
So that we can use commands like `nDMaterial ASDConcrete3D tag E v -fc fc -ft ft` without having to define hardening laws

### DIFF
--- a/SRC/material/nD/ASDConcrete3DMaterial.cpp
+++ b/SRC/material/nD/ASDConcrete3DMaterial.cpp
@@ -343,6 +343,29 @@ namespace {
 
 }
 
+// long name so not confused with other bezier3 functions
+double bezier3_for_ASDConcrete3D(double xi,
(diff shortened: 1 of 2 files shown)


MassimoPetracca:
@mhscott sounds good to me! I was also planning to add some simplified overloaded constructors to allow easier definitions of the model.
I would just move the bezier3_for_ASDConcrete3D into the anonymous namespace ( I typically put all utility functions there so that there will be no name conflict).
And the bezier3 declaration in the header file can be removed since it's not implemented in cpp.

mhscott:
@MassimoPetracca Thanks! I'll make those changes. I admit, I'm not super familiar with 21st century C++ -- `namespace`, `auto`, etc.  I originally was going to overload the constructor, but ran into some doubts about how to instantiate the hardening laws.

PR #1238, opened 2023-07-10
correct startTime implementation for PathSeries

mckee107:
Add startTime to PathSeries constructors. Add getStartTime() virtual function to TimeSeries and add default behavior to affected time series files.

For Tcl interpreter, users could specify `-startTime`, but the `PathSeries` constructors did not use the value.


mhscott:
@mckee107 Thanks! One suggestion - you can define the default `return 0.0;` behavior in the TimeSeries base class. Since the method is virtual, subclasses can override.

```cpp
virtual double getStartTime () {return 0.0;}
```

mhscott:
@mckee107 Some of the time series, e.g., `Pulse`, `Rectangular`, `Trig`, `Triangle`, already have start times as private data, but you return 0.0 for `getStartTime()`.

mckee107:
Interesting... I don't know how I overlooked that. Fixed it.

mckee107:
> I'm curious why you removed this `delete` statement.
> 
> image

It appears to be an innocent bystander from merging the upstream repo into my fork before submitting the pull request. I will go through the changes again to make sure nothing else was overlooked

PR #1239, opened 2023-07-10
add Ramp Series with smoothing option

mckee107:
Add Sigmoid-shaped ramp pattern. Documentation will be created in the documentation repo shortly.


mckee107:
Documentation pull request added ([#115](https://github.com/OpenSees/OpenSeesDocumentation/pull/115))

mckee107:
Currently, to my my knowledge, there is no time series function that can directly provide a smooth application of loads. For seismic analyses is not too important, but can be important for others. 

While this can be achieved with a path series script, it shouldn't have to be just because the path series function was there first. This is a non-invasive addition to the source code, which allows you to avoid unnecessary pre-processing and overhead. This function was first developed because of the amount of time and HPC allocations I've wasted dealing with the issues I've experienced using the path series function (differences in implementation for Tcl/PY, accuracy of documentation, etc.).

silviamazzoni:
what kind of issue do you encounter with HPC and the path series?

silviamazzoni:
Also, the Path series is universal and can accommodate all kinds of functions. The mode specialized code we put into OpenSees the heavier it gets, albeit in this case the code is "minimal."  It also makes learning OpenSees more complex.

mhscott:
There's a couple definitions for sigmoid or logistic functions. Perhaps more general with starting and stopping y-values, instead of [0,1]*factor?

@mckee107 Are you running analysis through log time? If I recall, this was an issue you had a while back.

silviamazzoni:
is there a way to make it ramp back to zero? that would be useful.

mckee107:
> Also, the Path series is universal and can accommodate all kinds of functions. The mode specialized code we put into OpenSees the heavier it gets, albeit in this case the code is "minimal." It also makes learning OpenSees more complex.

I agree, and as I dig through the source code to figure out why certain things happen the way they do, I run across *a lot* of undocumented features buried within. 

> I do think that this can be a valuable addition to the code because it provides a smooth-transition step function. Is there a way to make this even more general and maybe provide some examples of where it can be used to improve your model response? I think this would really help the users in deciding when it's beneficial to use it.

> There's a couple definitions for sigmoid or logistic functions. Perhaps more general with starting and stopping y-values, instead of [0,1]*factor?

> is there a way to make it ramp back to zero? that would be useful.

I can definitely work on a more general, easier to understand implementation. 


> I am, however, still curious about the issues you have encountered with the path series. are you using a file or sending values directly?

> Are you running analysis through log time? If I recall, this was an issue you had a while back.

I tried all of the different approaches using the PathSeries command, and I was still getting errors. I am running Life-Time analyses of RC structures using time-dependent models for aging (creep, shrinkage, corrosion), so I have a very large time-scale. At arbitrary points in time, a hazard load pattern is created and applied. So I can't use a time scale of years for the analysis, because dT can approach machine precision during the hazard analysis. And using seconds was causing issues integer overflows in the time series integrators, so that had to be addressed. The biggest thing was when trying to specify `-startTime` for the PathSeries (#1238), it just wasn't doing anything. I was getting seg faults because of the int overflows... but I couldn't set the time to 0 or change time scales because of the rate-dependent materials. These seg faults (and incorrect `-startTime` behavior) would happen when my models were towards the end of 18+ hour batch jobs on our HPRC clusters—a nightmare for debugging.

The `OPS_PatheSeries` function and the `TclSeriesCommand` parsing do not behave the same. And truthfully, I have stopped paying attention to the "old" documentation because it ends up contradicting the behavior I am seeing in the code. When trouble shooting, it becomes super difficult to determine where to start. Because of this, I look at the OpenSeesPY docs as a quick reference. I wrote a bunch of libraries for Tcl when I started my research in 2018, and there has never been a good time to sit down and convert everything to PY, so I make do.  So many functions have still use `Tcl_xyz_Command` and `OPS_xyz` parsers, and the OpenSeesPY group doesn't always mirror changes or features to the `Tcl_xyz_Command`.

mckee107:
I changed it to `RampSeries`. Takes the form of a linear ramp function with` tStart?` `tRamp?` and flags for `-offset offset?` `-smooth smoothFactor?` and `-factor cFactor?`

By indicating a smoothing factor (from 0 to 1), the start and end of the ramp are smoothed.

MWE:
```
 wipe
 model basic -ndm 2 -ndf 3;	    
 geomTransf	Corotational 1 ;
 
 after 1000
 set TestType EnergyIncr;
 set AlgType Newton
 set constType Transformation;
 set sysType FullGeneral;
 set numType RCM;
 set intType Newmark;

 node 1 0 0;
 node 2 0 120.0; 
 fix  1 1 1 1; 
  
 element elasticBeamColumn 1 1 2 484.0 4500.0 19521.3 1 

 # timeSeries Linear 1 
 timeSeries Ramp 1 10 90 -offset -5.0 -factor 2.0 -smooth 0.25

 pattern Plain 1 1 { 
 	load 2   0 -200.0 0;
 }
 
 recorder Node -file RBase.out -time -precision 15 -node 1 -dof 1 2 3 reaction
 recorder Pattern Ramp.out -time 1 
 
 test $TestType 1.0e-6 10 5;
 algorithm $AlgType ;
 constraints $constType; 
 system $sysType;
 numberer $numType;
 integrator $intType 0.5 0.25;
 analysis Transient ;
 
 puts "Gravity"
 for {set hh 0} {$hh < 100} {incr hh} {
	set ok [analyze 1 1.0 ]
	puts [format "Load Factor = %0.6f" [getLoadFactor 1]]
 }
 
 puts "Complete"
 
 wipe
 ```

mhscott:
Thanks, @mckee107! Please also update the documentation and the effects of `cFactor` and `offset` for negative values, e.g., to ramp to zero or ramp down from an initial value.

Issue #1255, opened 2023-08-06
Float32 data type used by the PVD recorder is sometimes too short.


gaaraujo:
Hello,

Someone in the OpenSees Facebook Group found an issue when trying to visualize the output of a PVD recorder in Paraview. See the original post [here](https://www.facebook.com/groups/opensees/permalink/10159738829816492/), along with a minimal working example provided by the user. 

The issue is that the PVD recorder sets the data type to "Float32" (see the source code for the [VTK](https://github.com/OpenSees/OpenSees/blob/96ba4cb9467afe7039c2ca86aa54a40f9950e281/SRC/recorder/VTK_Recorder.cpp#L650) and and [PVD](https://github.com/OpenSees/OpenSees/blob/96ba4cb9467afe7039c2ca86aa54a40f9950e281/SRC/recorder/PVDRecorder.cpp#L548) recorders), which cannot represent numbers smaller than 1e-38, while the exported data may include values in the order of 1e-200. Changing "Float32" to "Float64" inside the .VTU files generated by the recorder fixes the issue, but it can become annoying for larger models.

For clarity, I've attached a screenshot below, where you can see the mismatch between the data type and the order of magnitude of the output data.
![image](https://github.com/OpenSees/OpenSees/assets/24374005/5bb78140-dd86-4fed-b2ef-54224a65fdc1)

mhscott:
Thanks, make a pull request.

PR #1256, opened 2023-08-10
Fixing issue with data size and PVD recorder

gaaraujo:
I've changed all the occurrences of Int32 and Float32 to Int64 and Float64 in the PVD and VTK recorders, aiming at fixing issue #1255 .


mhscott:
@zhuminjie Is this PR OK?

PR #1258, opened 2023-08-12
Fix class name in error messages

lcpt:
I've simply replaced the correct class name: DisplacementPath in some error messages.

On the other hand, is not analysis/integrators the correct place for this class?


mhscott:
Probably should be moved, but does this class do anything different from DisplacementControl?

lcpt:
I'm not sure (the author was JZhong, not me).

The difference I've found is that this class uses an increment vector containing the increment to use in each step instead of the increment scalar used by the DisplacementControl class.

My intention was only to fix the error messages on this file to avoid confusion if someone uses it.

mhscott:
This integrator is not even in the Tcl or Python interpreters, so no one is using it, unless they are building C++ main functions. It would be helpful if you add the OPS commands and provide examples.

PR #1264, opened 2023-08-18
Implement `sendSelf` and `recvSelf` for `Pinching4`

ioannis-vm:
This PR implements the `sendSelf` and `recvSelf` methods of the `Pinching4` material.
As an initial attempt, I tried emulating the syntax in `ElasticMultiLinear` and `ElasticPowerFunc`, where `theChannel.<send/recv>Vector` is called repeatedly with different arguments, but I wasn't able to retrieve the correct vector values in `recvSelf` with this approach. The proposed implementation simply aggregates the individual values of the `Vector`-typed attributes inside a larger vector, skipping trial attributes.
I would be happy to also commit the `Tcl` script I used while debugging if that would be helpful.

### DIFF
--- a/SRC/actor/objectBroker/FEM_ObjectBrokerAllClasses.cpp
+++ b/SRC/actor/objectBroker/FEM_ObjectBrokerAllClasses.cpp
@@ -144,6 +144,7 @@
 #include "Hertzdamp.h"
 #include "JankowskiImpact.h"
 #include "ViscoelasticGap.h"
+#include "Pinching4Material.h"
 
 // Sections
 #include "ElasticSection2d.h"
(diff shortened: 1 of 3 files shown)


mhscott:
@ioannis-vm I think you want `int indx = 1;` because you are post-incrementing with `indx++`

ioannis-vm:
You are right. I updated the indexing. So, previously my implementation would overwrite the tag coming from `this->getTag()` and leave the last element unused. But then there is also `matTag`, which I included because it was in the private members in `Pinching4Material.h`. I tested a case where we don't transfer `matTag` and it appears to be working, but I was reluctant to exclude it. If it would be fine to do so, I can push another change.

mhscott:
You can remove `tagMat` from the private data... not sure why it was there in the first place because the material tag is stored in a parent class. And while you're at it, go ahead and remove the ```#ifdef _G3DEBUG``` blocks (the directives and the code they surround)

ioannis-vm:
I would be happy to make any other changes if needed.

PR #1280, opened 2023-09-10
Units  warning in limitCurve functions

L-iqra:
In the source code/files of ShearCurve.cpp, ShearCurve.h, RotationShearCurve.h, and RotationShearCurve.cpp, the units explicitly mentioned are psi and kips. However, the demonstration manual states that these functions can take input in both psi and MPa units. 

Link to [source codes](https://github.com/OpenSees/OpenSees/tree/master/SRC/material/uniaxial/limitState/limitCurve)
Link to [manual](https://opensees.berkeley.edu/wiki/images/5/57/LimitStateMaterialManual.pdf))

The users are advised to be cautious while using these functions.

@mhscott @fmckenna Can you please check this and make appropriate changes in the code ?

### DIFF
--- a/SRC/material/uniaxial/limitState/limitCurve/RotationShearCurve.cpp
+++ b/SRC/material/uniaxial/limitState/limitCurve/RotationShearCurve.cpp
@@ -57,6 +57,7 @@ static int shearCurveCount = 0;
 void *
 OPS_RotationShearCurve(void)
 {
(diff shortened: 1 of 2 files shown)


mhscott:
The manual, which was not written by the author of the LimitCurve classes, is incorrect? Instead, please update the documentation of OpenSees to be consistent.

L-iqra:
Thanks for the suggestion. I will close this PR.

PR #1282, opened 2023-09-15
Implement `sendSelf` and `recvSelf` for Steel4

ioannis-vm:
This pull request implements `sendSelf` and `recvSelf` for the `Steel4` uniaxial material. Similar to other uniaxial materials, the implementation gathers relevant variables in a single data array which is sent and then received.

The length of the vector-type private members of Steel4 is variable, with all such variables having the same length at every converged model state. This prevents directly sending those equal-length arrays. To overcome this, a second data array containing their size is also sent and received to inform the initalization of the data array in `recvSelf`, and everything else is transferred with a single data array.

### DIFF
--- a/SRC/material/uniaxial/Steel4.cpp
+++ b/SRC/material/uniaxial/Steel4.cpp
@@ -790,19 +790,312 @@ Steel4::revertToStart(void)
   }
 
   return 0;
+
 }
 
-int
-Steel4::sendSelf(int commitTag, Channel &theChannel)
-{
-  return -1;
+int Steel4::sendSelf(int commitTag, Channel &theChannel) {
+
+  // Obtain required vector size
+  int numPrevHalfCycles = dir_Par.size();
+
+  int vecSize = 0;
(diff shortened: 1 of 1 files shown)


mhscott:
@ioannis-vm Thanks! The vectors should be sent and received in the same order. Also, can you put the `numPrevHalfCycles` as the first vector entry and then send/recv only one vector instead of two?

ioannis-vm:
Thank you for the feedback. Indeed, it would be ideal to only have to transfer one vector, but I am not sure how `recvSelf` would know what size of a vector to initialize. Assuming `numPrevHalfCycles` is the first element of the vector, don't we still need to know the size a priori to receive the vector?

mhscott:
I see, you're right.

Another minor thing is you can send/recv `numPrevHalfCycles` in an ID (array of ints) object instead of Vector (array of doubles).

ioannis-vm:
That's a good point. I will work on another commit switching it to `ID`.

PR #1315, opened 2023-12-22
Adding MEFI element

carloslopezolea:
This PR adds the Membrane Fiber (MEFI) Element. 

The MEFI element is described by four nodes, each containing three degrees of freedom, two translations, and one in-plane rotation. The element formulation accommodates the quadrature points and weights of the classical finite element formulation of membrane elements to resemble strips (fibers), similarly to macroscopic elements.

Information about model formulation and validation can be found here https://www.sciencedirect.com/science/article/abs/pii/S0141029621018897


mhscott:
Add documentation at https://github.com/OpenSees/OpenSeesDocumentation

mhscott:
Why do you call this a "fiber" element? It is a 2D shell element that uses section pointers. The sections could be elastic, fiber, stress resultant, whatever, right? There's nothing in this element that pertains to fibers. `Shell2D` would be a more appropriate name.

mhscott:
Also, the PR is kinda useless if you're not including the section models this element works with. Three section forces, what are they?

carloslopezolea:
Hi Michael!

1. Documentation has been uploaded;
2. The MEFI element was developed for RC Walls based on a four-node membrane formulation with drilling DOF. The element formulation was adapted to include the concept of fibers (reinforced concrete panels; RCPanels), similarly to the SFI-MVLEM. Then, we can model a RC wall using only 1 MEFI element along the entire wall length (without kinematic assumptions);
3. Based on the previous point, the "fiber" concept is related to the element discretization about his length using RCPanels;
4. The element works with a "reinforced concrete layered membrane section". The person who is in charge of uploading the code of the section and new materials is working in some details about documentation. In a couple of days the section code is going to be uploaded. Sorry for the inconvenience. 

Regards,
Carlos.

mhscott:
With OpenSees, the term "fiber" is typically encapsulated within a beam or plate/shell section, not the element discretization. Your point 3 seems arbitrary. The element is a shell element, which is fine, but the name is misleading. In OpenSees the element and the section (constitutive response at integration points) are typically uncoupled. The notion of a "fiber element" is not accurate, especially in how you have implemented this element, which is a perfectly fine implementation, but the element implementation has nothing to do with whether or not there are fibers at the integration points.

fmckenna:
Happy New Year.  Just a gentle reminder that we are still waiting for the section updates. Also an example would be nice with the documentation so that we can test.

mhscott:
It would be nice if there's also an elastic section that could be used with this element, something akin to elastic sections in the nonlinear beam-column elements or elastic membrane sections for the 3D shells. Useful for verification and simple tests, which we are still waiting for. Thanks for adding the documentation and the sections and materials!

carloslopezolea:
Hi Michael!

1. I just updated the element documentation to include a pushover example for a RC wall that also performs a test comparing the expected and computed base shear.
2. We are going to work to add an elastic section that can be used with this element (probably in a couple of months).
3. In a short time, I'm going to upload the 3D version of the MEFI element (which includes an elastic plate for out-of-plane behavior, similar to the SFI-MVLEM-3D; MVLEM-3D; E-SFI-MVLEM-3D). Also, I'm planning to add the original quad element with rotational dof, from which the MEFI element is based.

Regards,
Carlos.

mhscott:
1. Thanks, but the expected and computed base shear should pretty much always be the same for a cantilever wall, right?
2. Why a couple of months? I'm only asking for an elastic section (E,nu,thickness) like ElasticMembranePlateSection.
3. MEFI is a membrane element with sections. Then when you add out of plane plate behavior, you will have a shell element. You should be able to handle the in-plane and out of plane response within the section and use existing shell elements. I am concerned because you will end up duplicating existing code.

carloslopezolea:
Michael,

1. Yes, I put a relative error of less than 1%, but it can be reduced considerably. Maybe 0.01% or less?
2. ReinforcedConcreteLayerMembraneSection02 is basically a wrapper that allows the use of any type of nDMaterial (not only FSAM as described in the documentation). Thus, we can use an elastic section as you requested.
Example: 
  nDMaterial ElasticIsotropic 1 $E $v
  nDMaterial PlaneStress 2 1
  section RCLMS02 3 2 $thickness

(I think the new sections are basically nDMaterials assemblers)

3. In MEFI-3D formulation the plate and membrane behaviors are uncoupled. (This simplification is commonly used for RC walls, as in SFI-MVLEM-3D, because the out-of-plane behavior is not very relevant and to reduce computational effort). Then, the plate implementation is included inside the MEFI-3D element code and not at section level. I believe this implementation is different from those available in OpenSees. Also, this element is very computationally efficient.

Regards,
Carlos.

mhscott:
Comparing base shear and applied loads is not very meaningful. Its equilibrium, can you also check displacements?

What is the difference between RCLMS01 and RCLMS02? Why do you call them "Reinforced Concrete" sections when they can use any nDMaterial? How about LayeredMembraneSection01 and 02? These classes sounds just like PlateFiberSection, but with only membrane stress resultants?

The fibers are in the section ("material assemblers" as you called them), not in the element. I still think you should reconsider the MEFI name so that users are not confused by inconsistent terminology.

carloslopezolea:
1. In the example, the pushover analysis was performed using a displacement-controlled analysis, so I considered it appropriate to compare the base shear with a known reference value. However, I can add a displacement comparison at nodes other than the one used for the displacement-controlled analysis. Is this correct?
2. RCLMS01 adds some special methods that are only appropriate for reinforced concrete materials (such as concrete cracking pattern). Probably RCLMS02 should be renamed to LayeredMembraneSection, we will check it.
3. Transversal fibers (along the element thickness) are defined at section level; however, longitudinal fibers (along the element length) are defined at element level. As MEFI is a membrane element, for a specific longitudinal fiber (i-th RCPanel), transversal fibers are evaluated for the same strain field. This element does not use the classical Gaussian quadrature rules, but instead uses the longitudinal fibers (RCPanels) to define the location of the quadrature points and weights. Also, this element has a published paper and another under review, in addition to references and presentations at international congresses, so I consider that it would not be appropriate to change the name.

Regards,
Carlos.

carloslopezolea:
Yes, you are right, the section stress resultant is Nx, Ny, and Nxy. We will correct it.

fmckenna:
just awaiting test with elastic section .. if we see that that works, we can take in the element as it demonstrates convergence problem is in the materials.

carloslopezolea:
Hi @fmckenna ,

Below is the same example from the documentation but using an elastic material

```tcl
# ------------------------------------------------------------------------------------------------------------------------------------------------------
# Start of model generation
# ------------------------------------------------------------------------------------------------------------------------------------------------------

wipe
model Basic -ndm 2 -ndf 3

# ------------------------------------------------------------------------------------------------------------------------------------------------------
# Set geometry, nodes, boundary conditions
# ------------------------------------------------------------------------------------------------------------------------------------------------------

# Nodes for RC wall
for {set i 1} {$i <= 8} {incr i} {
  node [expr 2*$i-1] 0.0    [expr (2209.8/7)*($i-1)]
  node [expr 2*$i]   1220.0 [expr (2209.8/7)*($i-1)]
}

# Nodes for loading transfer beam
for {set i 9} {$i <= 10} {incr i} {
  node [expr 2*$i-1] 0.0    [expr 2209.8 + (457.2/2)*($i-8)]
  node [expr 2*$i]   1220.0 [expr 2209.8 + (457.2/2)*($i-8)]
}

# Restraint fixes
fix 1 1 1 1                       
fix 2 1 1 1  

# Node restraints
equalDOF 15 16 1
equalDOF 17 18 1
equalDOF 19 20 1

# ------------------------------------------------------------------------------------------------------------------------------------------------------
# Define and build materials
# ------------------------------------------------------------------------------------------------------------------------------------------------------

set E 33234
set v 0.3
nDMaterial ElasticIsotropic 1 $E $v
nDMaterial PlaneStress 2 1
section LMS 3 152.4 1 -mat 2 -thick 152.4

# ------------------------------------------------------------------------------------------------------------------------------------------------------
# Define and build elements
# ------------------------------------------------------------------------------------------------------------------------------------------------------

# Build MEFI elements
set numFib 8;   # number of element macro-fibers
set db 228.6;   # boundary length discretization
set dw 127.133; # web length discretization

for {set i 1} {$i <= 9} {incr i} {
  element MEFI $i [expr 2*$i-1] [expr 2*$i] [expr 2*$i+2] [expr 2*$i+1] $numFib -width $db $dw $dw $dw $dw $dw $dw $db -sec 3 3 3 3 3 3 3 3;
}


# ------------------------------------------------------------------------------------------------------------------------------------------------------
# Gravity load analysis
# ------------------------------------------------------------------------------------------------------------------------------------------------------

# Define gravity load pattern
pattern Plain 1 Linear {
  load 19  0.0 -320272.0 0.0  
  load 20  0.0 -320272.0 0.0	
}

# Analysis generation
system BandGeneral
constraints Transformation
numberer RCM
test NormUnbalance 100.0 100 1
algorithm Newton
integrator LoadControl 0.05
analysis Static
set ok [analyze 20]
if {$ok == 0} {
 puts "Gravity analysis completed successfully";
} else {
 error "Gravity analysis failed";    
}
loadConst -time 0.0
wipeAnalysis

# ------------------------------------------------------------------------------------------------------------------------------------------------------
# Lateral load analysis
# ------------------------------------------------------------------------------------------------------------------------------------------------------

# Define lateral load pattern
pattern Plain 2 Linear {
  load 15 1000.0 0.0 0.0
  load 17 1000.0 0.0 0.0
  load 19 1000.0 0.0 0.0
}

# Analysis generation
system BandGeneral
constraints Transformation
numberer RCM
test NormDispIncr 0.001 100 1
algorithm Newton
integrator DisplacementControl 17 1 0.1
analysis Static
set ok [analyze 800]
if {$ok == 0} {
 puts "Pushover analysis completed successfully";
} else {
 error "Pushover analysis failed";    
}
```
The residual norms when using convergence test on displacement increment during pushover are shown below
image

Regards,
Carlos

mhscott:
Please resolve conflict, then we'll merge

carloslopezolea:
@mhscott conflicts were resolved!

PR #1319, opened 2023-12-30
Adding damping to Python interpreter

TangSEU:
Details see OpenSeesdampingCommands. CPP


mhscott:
Build fails.

Also, why do you do PRs by "file upload"? You should use command line or a git client, which are better approaches. There are many resources online for how to use git more efficiently.

mhscott:
For consistency, use capital `D` in the filename `OpenSeesDampingCommands.cpp`

PR #1338, opened 2024-01-29
Missing include in `HystereticSMMaterial.cpp`

claudioperez:
This material did not compile for me, `map` was not included.


mhscott:
Yeah, includes are messy in all the OpenSees classes, but bid the code not compile on a particular compiler or operating system? There have been no problems with this model on Mac, Linux, or Windows builds.

claudioperez:
Thanks for getting to this so fast! The compile failed on Arch Linux with GCC 13.2.1.

PR #1339, opened 2024-01-30
`LineSearch` leak in `NewtonLineSearch` ?

claudioperez:
It looks like the `NewtonLineSearch` algorithm is leaking its line search but I'm not sure. With the proposed change, the leak is gone and things seem to work well.


mhscott:
Thanks for catching that!

PR #1345, opened 2024-02-20
Run builds on pull request and add win64 vcxproj build

parched:
To ensure a pull request doesn't break the build, I've changed build to also run during all pull requests.
There are some GitHub settings that can be turned on to enforce this, which I recommend.

Also added here is a windows build using the vcxproj. I had no luck with the CMake build, but perhaps will investigate that further another time.


parched:
Hi @fmckenna @mhscott,

I've noticed that my PR has become outdated. Is this contribution still something you would like included in the repo? If so, I'll update it; otherwise, I'm happy to close it.

Thanks,
James

silviamazzoni:
This repo is missing the ASCE41 Strength degradation Files.

mhscott:
@parched Let's keep it open. If there are some updates, please push those changes.

parched:
@mhscott I've updated this now. However, I'm wondering if OpenSees actually wants to maintain the vcxproj files, instead of just the cmake files. If not, we could close this PR and I'll create another that deletes the whole Win64 directory.

Regardless, are the Win32 files planned to be maintained? I've updated a few in this PR, but they're still out of date. If they're not going to be maintained, I'll create another PR to delete them.

parched:
Closing this as there's now a cmake build. IMO we should remove all the other build systems to reduce confusion.

PR #1346, opened 2024-02-21
Introduce `ModElasticBeam3D` element

ioannis-vm:
This commit introduces the 3D counterpart to the `ModElasticBeam2D` element. This element enables utilizing the concentrated plasticity modeling approach discussed in the [OpenSees wiki](https://opensees.berkeley.edu/wiki/index.php/Pushover_Analysis_of_2-Story_Moment_Frame) on the _"Pushover Analysis of 2-Story Moment Frame"_ with models defined in 3D. If directed to do so, I would be happy to work on documentation or any other further addition/development that could make this element more useful to the user base.


mhscott:
Indices should be 3,4 on kb in `getInitialStiff` method for bending about y. Correct in `getTangentStiff`

ioannis-vm:
Thank you for reviewing the code, and thanks for noticing that.

mhscott:
I know there's not a documentation page for `ModElasticBeam2d` at [OpenSeesDocumentation](https://github.com/OpenSees/OpenSeesDocumentation), but would you be willing to add documentation for these elements? 2d and 3d are virtually the same.

ioannis-vm:
I firmly believe that good documentation greatly enhances the quality of open-source projects and is very helpful to the users, so I would be happy to contribute. Would you like to give me a brief summary of what you 'd like me to include in the docs? I imagine we would need a separate entry for each element. We can include a link to the wiki page I shared above as a note, for further reading, and possibly also add some references on the lumped plasticity approach these models are intended to add support for. Othar than these (and obviously the commands and their arguments), what else do you think would benefit those entries? How soon would you need that PR?

mhscott:
Thanks! Something basic like what's done with the `elasticBeamColumn` would be good.
https://github.com/OpenSees/OpenSeesDocumentation/blob/master/source/user/manual/model/elements/elasticBeamColumn.rst
The page has both 2d and 3d command inputs.
You can also add references to other sources.

PR #1353, opened 2024-03-01
Remove MySQL as a required package

mckee107:
It doesn't seem necessary to have MySQL included as a required package, this pull request changes the `REQUIRED` dependency type from the `find_package(MySQL ...)`  command. 

Requiring `MySQL` is a headache when building OpenSees within the structure of the `EasyBuild` software build and installation framework used on our HPC clusters. 

If it can be removed as a requirement, it would solve a lot of issues. May need to add  `_HAVE_MYSQL` flags in other locations, but all checks pass without them at the moment.


fmckenna:
It may need to go back if I ever decide we want to include the mysql database.

PR #1354, opened 2024-03-04
Some files relying on `OPS_Globals.h` for `string.h`.

claudioperez:


fmckenna:
Claudio .. some of files already have #include <string> .. if <string> not needed, can you remove .. also just curios as to what compiler is flagging these

PR #1358, opened 2024-03-12
Correct term in coefficient series expansion for corotational shells.

claudioperez:
It appears the factor $\frac{1}{270}$ in the modified lines is wrong. I believe the relevant expression is Equation (101) in https://doi.org/10.1016/j.cma.2004.07.035:

$$
\begin{aligned}
\mathbf{H}(\boldsymbol{\theta}) & =\frac{\partial \boldsymbol{\theta}}{\partial \boldsymbol{\omega}}=\mathbf{I}-\frac{1}{2} \boldsymbol{\Theta}+\eta \boldsymbol{\Theta}^2 \quad \text { with } \eta=\frac{1-\frac{1}{2} \theta \cot \left(\frac{1}{2} \theta\right)}{\theta^2}  =\frac{1}{12}+\frac{1}{720} \theta^2+\frac{1}{30240} \theta^4+\frac{1}{1209600} \theta^6+\cdots
\end{aligned}
$$

Here is a short Python script that shows using $\frac{1}{720}$ is indeed closer to the closed form expression:
```python
import numpy as np
from numpy import tan

cot = lambda x: 1/tan(x)

def test(theta):
  closed_a = (1-0.5*theta*cot(0.5*theta))/theta**2;
  closed_b = (1.0 - 0.5 * theta * tan(0.5 * np.pi - 0.5 * theta)) / (theta * theta)
  
  # expression used in pull request:
  series_a = 1/12 + theta**2/720 + theta**4/30240 + theta**6/1209600;
  # expression currently in use:
  series_b = 1/12 + theta**2/270 + theta**4/30240 + theta**6/1209600;
  
  print(f"{closed_a - series_a = }")
  print(f"{closed_a - series_b = }")

  print(f"{closed_b - series_a = }")
  print(f"{closed_b - series_b = }")


test(0.1)
# closed_a - series_a = 1.096345236817342e-15
# closed_a - series_b = -2.3148148147064673e-05
# closed_b - series_a = 2.1203871991559708e-13
# closed_b - series_b = -2.31481479361223e-05

test(0.01)
# closed_a - series_a = 2.267352972040726e-13
# closed_a - series_b = -2.3148125474770076e-07
# closed_b - series_a = -9.081155272205876e-11
# closed_b - series_b = -2.3157229303572002e-07
```

### DIFF
--- a/SRC/element/shell/ASDEICR.h
+++ b/SRC/element/shell/ASDEICR.h
@@ -378,13 +378,14 @@ public:
 
 			if (angle >= 2.0 * M_PI)
 				angle = std::fmod(angle, 2.0 * M_PI);
-
+			
+                        // compute eta from Felippa et al 2005, Equation (101)
 			double eta;
 			if (angle < 0.05) {
 				double angle2 = angle * angle;
 				double angle4 = angle2 * angle2;
 				double angle6 = angle4 * angle2;
-				eta = 1.0 / 12.0 + 1.0 / 270.0 * angle2 + 1.0 / 30240.0 * angle4 + 1.0 / 1209600.0 * angle6;
+				eta = 1.0 / 12.0 + 1.0 / 720.0 * angle2 + 1.0 / 30240.0 * angle4 + 1.0 / 1209600.0 * angle6;
 			}
 			else {
 				eta = (1.0 - 0.5 * angle * std::tan(0.5 * M_PI - 0.5 * angle)) / (angle * angle);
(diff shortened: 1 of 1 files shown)


fmckenna:
@MassimoPetracca can you review this pull request. thanks

fmckenna:
@MassimoPetracca a gentle ping .. Filip reviewed but wanting to make sure you agree as it migh effect some of results in your verification suite

MassimoPetracca:
@claudioperez thanks for spotting it!
@fmckenna @mhscott sorry for the late response but I missed this PR...
Anyways, there is no problem regarding the invalidation of results because 1) that term is very small and is an approximation used only when the angle is small, and 2) it appears only in the LHS, so at most it can improve convergence. But the RHS is not affected by this.
So good for me, thanks!

PR #1364, opened 2024-03-16
Update c make list glob method

mckee107:
The current method hard-codes directories and sends them to `include_directories(...)` AND THEN uses a recursive glob method to collect the same directories and send them (again) to `include_directories(...)`. This causes the linker command line to be massively long and causes `command line too long` errors.

Additionally, using `CMakeList.txt` as the search term rather than `*.h` is the preferred method. This ensures that only directories that can actually be built/included with CMake are included.... No CMakeList.txt, no CMake build.

` ${OPS_BUNDLED_DIR}/CMakeLists.txt` has been commented, since this line is likely the root of a lot of compile issues that occur. This line was adding all of the `/OTHER` directories to the `include_directories(...)` which leads to conflicts during linking when there are multiple versions of the same package and/or user-supplied external libraries

Since `/OTHER` directories were removed from the recursive glob search method, the hard-coded `/OTHER` directories must also be sent to `include_directories(...)`
Ideally, there should be no hard-coded paths (especially to libraries) in the CMakeList... These should be added as a configuration `option(...)` with the default value pointing towards the `/OTHER` location. This would give the user the opportunity to supply alternate locations (`-D<LIB>_INCLUDE_DIR` , `-D<LIB>_LIBRARIES`) without needing to patch or revise the CMakeList first...I can tackle this soon

### DIFF
--- a/SRC/api/CMakeLists.txt
+++ b/SRC/api/CMakeLists.txt
@@ -30,4 +30,4 @@ target_sources(OPS_Api PUBLIC
 endif()
 
 target_include_directories(OPS_Api PUBLIC ${CMAKE_CURRENT_LIST_DIR})
-
+set_target_properties(OPS_Api PROPERTIES LINKER_LANGUAGE CXX)
--- a/SRC/java/CMakeLists.txt
+++ b/SRC/java/CMakeLists.txt
@@ -14,4 +14,4 @@ target_sources(OPS_Java
 )
 target_include_directories(OPS_Java PUBLIC ${CMAKE_CURRENT_LIST_DIR})
 
-add_subdirectory(./kepler/)
+add_subdirectory(kepler)
--- a/SRC/material/section/repres/CMakeLists.txt
+++ b/SRC/material/section/repres/CMakeLists.txt
@@ -8,8 +8,8 @@ add_library(OPS_Section_Repres OBJECT)
 
 target_include_directories(OPS_Section_Repres PUBLIC ${CMAKE_CURRENT_LIST_DIR})
 
-add_subdirectory(./cell/)
-add_subdirectory(./patch/)
-add_subdirectory(./reinfBar/)
-add_subdirectory(./reinfLayer/)
-add_subdirectory(./section/)
+add_subdirectory(cell)
+add_subdirectory(patch)
+add_subdirectory(reinfBar)
+add_subdirectory(reinfLayer)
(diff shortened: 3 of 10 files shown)


fmckenna:
this pull request is breaking the builds. can you fix it.

CMakeLists.txt
```diff
@@ -184,68 +184,12 @@ set(CMAKE_CXX_ARCHIVE_CREATE "<CMAKE_AR> cqls <TARGET> <LINK_FLAGS> <OBJECTS>")
...
-
-# TODO: Temporary fix; include all directories with .h files through glob
+# include all directories with CMakeLists.txt files through glob
+# Searching for *.h masks when CMakeLists.txt are not setup
+include_directories(${OPS_SRC_DIR})
 file(GLOB_RECURSE ops_include_files CONFIGURE_DEPENDS 
-    ${OPS_SRC_DIR}/*.h
-    ${OPS_BUNDLED_DIR}/*.h
+    ${OPS_SRC_DIR}/CMakeLists.txt
+    #${OPS_BUNDLED_DIR}/CMakeLists.txt
```
claudioperez, line 192: This change does not make sense to me. Why are CMake files being globbed for includes? If the glob results are passed to `include_directories()`, shouldn't they be header files?

I did the globbing as a temporary patch and it should eventually be removed, as it is generally poor practice and doesn't let the build system make a proper dependency representation. Hard-coded includes are the best option for the core subdirectories. I've done this on my fork (https://github.com/claudioperez/OpenSeesRT/blob/master/CMakeLists.txt) which has significantly improved build quality, but I have not had a chance to merge it in.

However, for external libraries, it would be great to have user-configurable paths. I originally implemented a system for this, but things got messy and we fell back to a combination of `find_package` and/or Conan. I'd love to see some progress on this.

claudioperez:
I did the globbing as a temporary patch and it should eventually be removed, as it is generally poor practice and doesn't let the build system make a proper dependency representation. Hard-coded includes are the best option for the core subdirectories. I've done this on my fork (https://github.com/claudioperez/OpenSeesRT/blob/master/CMakeLists.txt), which has significantly improved build quality, but I have not had a chance to merge it in.

However, for external libraries, it would be great to have user-configurable paths. I originally implemented a system for this, but things got messy and we fell back to a combination of `find_package` and/or Conan. I'd love to see some progress on this.

PR #1367, opened 2024-03-23
TriSurfaceLoad - add to Python interpreter, update normalization

mhscott:
This analysis with one `TriSurfaceLoad` should give reactions that equal pressure * element area.

Am I missing something, @jaabell, or was there an issue when you copied `SurfaceLoad` to `TriSurfaceLoad`?

```python
ops.wipe()
ops.model('basic','-ndm',3,'-ndf',3)

ops.node(1,0,0,0); ops.fix(1,1,1,1)
ops.node(2,1,-1,0); ops.fix(2,1,1,1)
ops.node(3,1,1,0); ops.fix(3,1,1,1)

p = 1

ops.element('TriSurfaceLoad',3,1,2,3,p)

ops.system('UmfPack')
ops.analysis('Static','-noWarnings')

ops.analyze(1)
ops.reactions()

for nd in ops.getNodeTags():
    print(ops.nodeReaction(nd))
```

### DIFF
--- a/SRC/element/surfaceLoad/TriSurfaceLoad.cpp
+++ b/SRC/element/surfaceLoad/TriSurfaceLoad.cpp
@@ -244,7 +244,7 @@ TriSurfaceLoad::UpdateBase(double Xi, double Eta)
 
     // Normalize
     // double norm = myNhat.Norm();
-    myNhat = myNhat / 2;
+    myNhat = myNhat / 3;
 
     return 0;
 }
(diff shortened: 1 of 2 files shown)


jaabell:
Hi Michael. Seems like you figured it out. Right?

mhscott:
@jaabell Did I fix it "correctly"? Better to divide `myNhat` by 3 and leave `myNI` all 0.5 -or- to divide `myNhat` by 2 and make `myNI` all 0.3333333 instead of 0.5?

jaabell:
Hello Michael, 

Sorry this took so long. 

You did fix it correctly, in the sense that it now gives the correct result. However, I think your second fix suggestion makes more sense, because `myNhat `comes from the cross product of `g1 `and `g2 ` which is proportional to element area, hence the division by two, otherwise you get the area of the parallelepiped defined by `g1 `and `g2`. My original mistake was, as you spotted, to not update the definition of the shape functions at the centroid of the triangle. I'll PR a refactor of the code to reflect this change. 

Thanks!
JAA

jaabell:
Refactor done: #1399

PR #1372, opened 2024-03-25
Init disp fix

jaabell:
Sets the default behavior of the tetrahedron elements to NOT consider initial displacements, as per suggestion by @mhscott. Use the new <doInitDisp?> flag to enable the accounting of initial displacements when setDomain is called.


fmckenna:
Jose,

could yoou redo the optional init disp to be more consistant .. something like 
int doInit= 0;
```
while (OPS_GetNumRemainingInputArgs() > 0) {
        type = OPS_GetString();
        if (strcmp(type,"-doInitDisp") == 0) {
            doInit = 1;
        } 
       ... blah blah blah
}
```

mhscott:
Like `'-doRayleigh',1` or `0` for zeroLength elements. Makes it easier to parse optional arguments as keywords instead of positional.

jaabell:
Done!

PR #1377, opened 2024-03-31
Some warning fixes

claudioperez:


SRC/element/dispBeamColumn/DispBeamColumnNL3d.cpp
```diff
@@ -2343,10 +2343,9 @@ DispBeamColumnNL3d::getResistingForceSensitivity(int gradNumber)
 
       continue;
 
-      double si;
       for(int j = 0; j < order; j++) {
+	double si = s(j)*wt[i];
 	switch(code(j)) {
-	  si = s(j)*wt[i];
```
claudioperez, line 2349: This is the simplest possible fix, but it may be best to just remove everthing that follows `continue` to keep things clean.

mhscott: You can block comment the code, but don't remove.

Did you also fix the warning for `cmass` in the OPS function?

claudioperez: I havent seen that warning, but I'm compiling from a branch that is missing some of the OPS functions.

claudioperez:
Can `currentColID` be removed here? https://github.com/claudioperez/OpenSees-Upstream/blob/da3d9e08051cc3f539f5d1a775a9166c9fe194e8/SRC/handler/DataFileStreamAdd.cpp#L348-L351

PR #1388, opened 2024-04-17
An asynchronous parallel double-stage viscous fluid damper model

buceaxll:
Incorporated an uniaxialMaterial model that simulates the hysteretic responses (axial load-deformation) of an asynchronous parallel double-stage viscous fluid damper (APDVFD) with specific parameters.


buceaxll:
Incorporated an uniaxialMaterial model that simulates the hysteretic responses (axial load-deformation) of an asynchronous parallel double-stage metallic damper (APDMD) with specific parameters.

PR #1405, opened 2024-04-26
Fixing getCopy() in SLModel

mhscott:
@Diegohere Please review. Also, how in the world was this version of the model - with such a major error in `getCopy()` - verified, validated, and used to generate (presumably) and/or compare with publishable results? I do not understand.

### DIFF
--- a/SRC/material/uniaxial/SLModel.cpp
+++ b/SRC/material/uniaxial/SLModel.cpp
@@ -918,7 +918,7 @@ SLModel::revertToStart(void)
 UniaxialMaterial *
 SLModel::getCopy(void)
 {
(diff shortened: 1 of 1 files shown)


Diegohere (approved):

mhscott:
@Diegohere Thank you for reviewing and approving.

I'm still curious to hear an explanation as to how this version is what was offered up for public OpenSees. This mistake would cause the `SLModel` to *never* work unless sigmaY and E are both 1.0.

Diegohere:
I had written and tested the version with the mistake with OpenSees V3.3.0 and never got any problems. I have used the material law within a force-based beam-column element.

mhscott:
I don't see how that's possible based on the commit history, unless your version 3.3.0 was compiled from a different repository.

Anyway, I'll drop it. Thanks for looking over the changes.

PR #1407, opened 2024-04-26
Activate Beam3dPartialUniformLoad function for 3d ForceBeamColumn element

tpanagiotakos:


mhscott:
Were the signs on shear incorrect, or did you change the signs to match your preference?

tpanagiotakos:
The change of signs is necessary because otherwise, the result obtained in the section is incorrect. The change is made because the shear Vz must be compatible with the moment My and similarly the shear Vy with the moment Mz."

mhscott:
OK, I see. Thanks!

PR #1434, opened 2024-05-22
updating VS2022 project files for win32 and win64

aschellenberg74:


mhscott:
Can you set the toolset back to v142 instead of v143?

aschellenberg74:
I have reverted the toolset to v142.

PR #1448, opened 2024-06-13
Compute shear strain and shear force in the panel

gtorrisi77:
Compute shear strain and shear force in the panel


gtorrisi77:
I don't know why all ckecks have failed, I have only add some lines to an existing working file

mhscott:
The lines you added have syntax errors. Did you compile locally before making the PR?

gtorrisi77:
> The lines you added have syntax errors. Did you compile locally before making the PR?

Yes! I have compiled and tested the file, but I uploaded the wrong file (problem with my local network).  Now I updated the repo with the correct file.

PR #1463, opened 2024-07-02
Conan 2.0

z-ichinohe:
conanfile.py and Github Action workflows are revised for Conan 2.0.
It may include updates on building procedure. OpenSeesDocumentation is also updated in OpenSees/OpenSeesDocumentation.


fmckenna:
we moved to conan2 .. thanks

PR #1469, opened 2024-07-19
Add runtime

claudioperez:
A new runtime framework and interpreter are implemented that reproduce and extend both the existing Tcl and OpenSeesPy interpreters. Global variables are eliminated and program state is encapsulated in two "runtime" classes. The resulting codebase is:
- easy to reason about, 
- easy to maintain, and
- eliminates the need to maintain two distinct interpreter APIs.

This pull request will not affect the existing interpreters. **No files are modified, only new files are added**.

## Design

- **Consolidated Containers** The type-safe [`getTypedObject<>`](https://github.com/claudioperez/OpenSees-Upstream/blob/4684719f16cf7ae49a355b9738a612e83dcebdad/SRC/runtime/runtime/BasicModelBuilder.h#L84)  template method and friends eliminate the need for several global container variables scattered throughout the codebase and consolidates tagged object storage. See, for example: [`section.cpp`](https://github.com/claudioperez/OpenSees-Upstream/blob/8a054eb0cc2d3532a697a115ec92805eecbf579b/SRC/runtime/commands/modeling/section.cpp#L968):
   ```c++
   torsion = builder->getTypedObject<UniaxialMaterial>(torsionTag);
   ```
 
- **Compile-time Extensibility** User models are declared in a single declarative initializer and require no explicit runtime function calls (like `std::insert`/`std::pair`) or modifications to control flow structures (like extending an `if () {} else {}` chain). For example, [`material.hpp`](https://github.com/claudioperez/OpenSees-Upstream/blob/4684719f16cf7ae49a355b9738a612e83dcebdad/SRC/runtime/commands/modeling/material.hpp#L74-L81)
   ```c++   
  static std::unordered_map<std::string, OPS_Routine*> material_dispatch = {
    {"InitStressMaterial",            OPS_InitStressNDMaterial},
    {"InitStrainMaterial",            OPS_InitStrainNDMaterial},
    {"InitStrain",                    OPS_InitStrainNDMaterial},
  
    {"ReinforcedConcretePlaneStress", OPS_ReinforcedConcretePlaneStressMaterial},
    {"PlaneStressLayeredMaterial",    OPS_PlaneStressLayeredMaterial},
    {"PlaneStressRebarMaterial",      OPS_PlaneStressRebarMaterial},
   ```

- **Robust** All objects are tightly encapsulated. Commands can only alter the analysis objects if they are explicitly passed the analysis runtime via the client data argument. These commands are all implemented under `commands/analysis`. The same goes for the model builder, with commands under `commands/modeling`. This greatly reduces the propensity for invalid access to global pointers.

- **Idiomatic Parsing** All commands are implemented by parsing an `argv`-like array. Modeling and analysis state is passed explicitly by an extra "client data" parameter. Motivation for this includes:
  - **Standard practice** The majority of interpreters that support implementing user commands from C do so by supplying the user with an array of interpreter objects. See, eg, the standard `main()` function from C/C++ and the C interfaces for MATLAB, Python, Tcl, Lua, etc.  This way users will be able to employ familiar idioms.
  - **Maintenance** Parsing procedures that employ the `OPS_Get*` functions are difficult to reason about and require programmers to mentally keep track of the parser's implicit state.
  - **Thread-safety** Because parsing functions harbor no implicit state, everything is inherently thread-safe and parallelizable.

## User-Facing Features

- new `surface` command in Python. Works like `block2D`, but:
  - Does not require hard-coding element types
  - Does not impose a fixed number of arguments for element types. The way `block2D` is implemented in OpenSeesPy, it is impossible to pass optional arguments to elements. The new `surface` method is introduced because `block2D` probably can't be fixed for backwards-compatibility reasons.
  - Optionally joins with existing mesh, eliminating the need for slow, cumbersome, and error prone constraints. See example [here](https://stairlab.github.io/opensees-gallery/examples/wrench/)

- new `printA` features:
  - does not require `FullGen` solver; now all one needs to do is call `printA`. This is possible due to the new consolidated memory ownership model.

  - new `-m, -c, -k` options eliminate need for manually changing to `GimmeMCK` integrator. To
    form a linear combination of mass and stiffness just run:
    ```tcl
    printA -m 0.5 -k 0.1
    ```
    in Tcl, or 
    ```python
    M = model.getTangent(m=0.5, k=0.1)
    ```
    in Python. An instance of `GimmeMCK` is implicitly created and destroyed, and any previously used integrator is restored. See the implementation [here](https://github.com/claudioperez/OpenSees-Upstream/blob/4684719f16cf7ae49a355b9738a612e83dcebdad/SRC/runtime/commands/analysis/analysis.cpp#L515)

  - new `getTangent()` method in Python eliminates need to pass `-ret` flag and follows standard Python semantics.

- add `section` argument to `fiber`, `patch` and `layer` commands in Python

- add `pattern` argument to `load` command

- new `invoke` Tcl command and Python constructs allow users to directly invoke object methods without switching between dedicated "testing" model builders. For example, to set the strain in a `UniaxialMaterial` with tag `1` to `0.01`, run:
  ```tcl
  invoke UniaxialMaterial 1 {strain 0.01}
  ```
  To print the stress:
  ```tcl
  puts [invoke UniaxialMaterial 1 {stress}]
  ```
  or, all together:
  ```tcl
  puts [invoke UniaxialMaterial 1 {strain 0.01; stress}]
  ```

- new `progress` command in Tcl provides a simple way to display a progress bar. This is usable from Python, but should not be marketed as there are several better options for this kind of thing in Python.


claudioperez:
Please let me know if there is anything needed before this can be merged.

claudioperez:
This PR now includes a fix for https://github.com/OpenSees/OpenSees/issues/1494 in the Tcl interpreter

mhscott:
Please make the simple bug fixes in a separate branch.

claudioperez:
Thanks for looking it over @mhscott . I opened a separate pull request at https://github.com/OpenSees/OpenSees/pull/1496 . Is there anything needed for this one to be merged?

claudioperez:
Please let me know if you have any suggestions.

PR #1478, opened 2024-08-02
Removes Andes Shell, H5DRM, ASDPlastic

jaabell:


jaabell:
The original PR removes the old verions of the code. 
Now there are new commits to add back H5DRMLoadPattern and ASDPlasticMaterial3D. 

I will not be putting the ANDES Shell back in.

Issue #1501, opened 2024-09-06
Compilation errors Windows


gaaraujo:
Hello,

I am trying to compile OpenSees following the instructions provided in the [OpenSees Documentation](https://opensees.github.io/OpenSeesDocumentation/developer/build.html#windows-10). When trying to run the following line in my cmd terminal, I get LINK errors:

```
cmake --build . --config Release --target OpenSees --parallel 4
```

The errors I get look like this:

```
48>OPS_Element.vcxproj -> C:\Users\garaujor\source\repos\OpenSees\build\OPS_Element.dir\Release\OPS_Element.lib
48>Done building project "OPS_Element.vcxproj".
51>C:\Users\garaujor\source\repos\OpenSees\SRC\material\nD\ASDPlasticMaterial3D\ElasticityModels\DuncanChang_EL.h(61,56): error C2065: 'M_PI': undeclared identifier
51>C:\Users\garaujor\source\repos\OpenSees\SRC\material\nD\ASDPlasticMaterial3D\ASDPlasticMaterial3D.h(484,32): message : see reference to function template instantiation 'const VoigtMatrix &DuncanChang_EL::operator ()<utuple_storage<std::tuple<Dilatancy,ReferenceYoungsModulus,PoissonsRatio,ReferencePressure,DuncanChang_MaxSigma3,DuncanChang_n,TensorLinearHardeningParameter,ScalarLinearHardeningParameter,T,ModelParameterType<double,AF_cr_Name>,MassDensity,InitialP0>>>(const VoigtVector &,const ParameterStorageType &) const' being compiled
51>        with
51>        [
51>            T=AF_ha,
51>            ParameterStorageType=utuple_storage<std::tuple<Dilatancy,ReferenceYoungsModulus,PoissonsRatio,ReferencePressure,DuncanChang_MaxSigma3,DuncanChang_n,TensorLinearHardeningParameter,ScalarLinearHardeningParameter,AF_ha,AF_cr,MassDensity,InitialP0>>
51>        ]
51>C:\Users\garaujor\source\repos\OpenSees\SRC\material\nD\ASDPlasticMaterial3D\ASDPlasticMaterial3D.h(480,19): message : while compiling class template member function 'const Matrix &ASDPlasticMaterial3D<EL,YF,PF,10000>::getInitialTangent(void)'
51>        with
51>        [
51>            EL=DuncanChang_EL,
[...]

PR #1506, opened 2024-09-06
Adding definition for M_PI in DuncanChang_EL.h. Fixes compilation error on Windows

gaaraujo:
M_PI was not defined in DuncanChang_EL.h. This would prevent compilation on Windows, as mentioned in issue #1501 .


gaaraujo:
@jaabell, does this change seem appropriate, or do you think there’s a better file for defining `M_PI`? I’m wondering if it should be moved elsewhere to keep things consistent with the rest of the files in [`SRC/material/nD/ASDPlasticMaterial3D`](https://github.com/OpenSees/OpenSees/tree/f3397a4425f1b89a1a2317a1694b824dbf65ff5b/SRC/material/nD/ASDPlasticMaterial3D).

fmckenna:
we should probably put something in OPS_Globals.h

jaabell:
> M_PI was not defined in DuncanChang_EL.h. This would prevent compilation on Windows, as mentioned in issue #1501 .

Thank you

PR #1516, opened 2024-09-18
cmp - implement `-dof` switch in `fix` (ready)

claudioperez:
This PR allows the `fix` command to be used for single DOF at a time. This was implemented to streamline the implementation of an OpenSees frontend which needs to fix individual DOFs at a time, but it is expected that general users will find this useful.

To fix DOF with index `dof` at a node with tag `node` the syntax in Tcl is:
```tcl
fix $node -dof $dof
```
for Python with OpenSeesRT:
```python
opensees.openseespy.fix(node, dof=dof)
```
or using the `opensees.openseespy.Model` class:
```python
model.fix(node, dof=dof)
```


mhscott:
Please add this to the standard `fix` command too.

claudioperez:
@mhscott Thanks for taking a look, I've just added it to the standard command. I can try adding it to OpenSeesPy too, I'll get back to you with any progress.

PR #1534, opened 2024-10-15
cmp - fix broken JSON output

claudioperez:
- JSON output has been added to `Steel01Thermal`. 
- `FiberSection3dThermal::Print` wasn't handling the print flag correctly and was producing non-JSON output, even when JSON was requested.
- `FiberSection3dThermal` incorrectly identified itself as `"FiberSection3d"`. The `getClassType()` method is used to prevent similar copy-paste errors in the future.
- Constants are defined that eliminate the need for hard-coding indentation depths in JSON output. Two constants are added that refer to the `"properties"` and `"geometry"` sections of the JSON output, respectively. The constants are implemented as protected constant-expression static members of `TaggedObject`, which defines the `Print()` abstract method.

### DIFF
--- a/SRC/material/section/FiberSection3dThermal.cpp
+++ b/SRC/material/section/FiberSection3dThermal.cpp
@@ -1222,63 +1222,66 @@ void
 FiberSection3dThermal::Print(OPS_Stream &s, int flag)
 {
   if (flag == 2) {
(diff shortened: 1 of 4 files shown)


claudioperez:
Indentation constants have been taken out of `TaggedObject` and moved into the `OpenSees::Printing` namespace

PR #1559, opened 2024-12-27
element api fix, re-enabling ops_TheActiveDomain

johnnyontheweb:
Fixing issue in element API by re-enabling ops_TheActiveDomain.
Actually, all the functions related to the domain returned 0.
This is also generally required to compile against fortran elements.


fmckenna:
We do not recomend using the dynamic linkiung approach any more due our preference for making people use Cmake, I will accept BUT do not gaurantee long term it will be there.

PR #1570, opened 2025-02-05
OPS_PathSeries update

zhuminjie:


gaaraujo:
Not really a big issue, but please note that this PR changes how the `Path` timeSeries interprets list inputs in Tcl scripts.
Previously, you could define a list of values and use it to create a `Path` timeSeries like this:
```tcl 
set values {0.1 0.2 -0.5}; # or even [list 0.1 0.2 -0.5]
timeSeries Path 13 -dt 0.0025 -values $values 
```
and it would work.

Now, the interpreter expects an expanded list of values. Therefore, you need to write:
```tcl
timeSeries Path 13 -dt 0.0025 -values {*}$values
```
Otherwise, you will encounter an error like:
```stdout
WARNING choice of options for path series is invalid
expected floating-point number but got "0.1 0.2 -0.5"
```

PR #1572, opened 2025-02-13
Build OpenSees on Mac with arm64 arch

vot4anto:
add folder Portfile with instruction on how use MacPorts to build OpenSees on Mac arm64


vot4anto:
It is enough to set a env variable Eigen3_INCLUDE_DIR and use brew or port as prefered

PR #1583, opened 2025-03-04
New element: ZeroLengthNew

XianjueDeng:
A new element is developed: ZeroLengthNew. This element fixed multiple bugs in the zeroLengthInterface2D and a new feature was added.

Bugs fixed in zeroLengthInterface2D: (detailed descriptions are given at the end)
1. Incorrect friction behavior when the interface is under cyclic sliding motion.
2. Incorrect normal pressure behavior due to the repetitive search algorithm. zeroLengthInterface2D will overestimate normal pressure at the interface when sliding occurs.

New features:
1. Allow user to change friction coefficient during the analysis.

Known issue:
1. The ZeroLengthNew element properly captured the behavior when the two sides under contact separated (detach). However, detach introduces sharp changes in normal pressure, friction, and internal variables in the interface and may results in numerical difficulties and long analysis time.

User Instruction:
1. The tcl input command for ZeroLengthNew element is the same as zerolengthInterface2D:
**element ZeroLengthNew eleTag? -sNdNum sNdNum? -pNdNum pNdNum? –dof sdof? mdof? -Nodes Nodes? Kn? Kt? phi?**
2. The following is a tcl input example:
image
**element ZeroLengthNew 1 -sNdNum 6 -mNdNum 2 -dof 3 2 -Nodes 10 9 8 7 6 5 4 1 $Kn $Kt $phi**
_(the above defines ZeroLengthNew element between quad element in blue and beam elements in red)_
4. To change the friction coefficient during the analysis, update the parameter named "phi". The following is an example updating the friction coefficient:
```
set Kn 1e5
set Kt 1e5
set phi_ini 0.0
element ZeroLengthNew 1 -sNdNum 6 -mNdNum 2 -dof 3 2 -Nodes 10 9 8 7 6 5 4 1 $Kn $Kt $phi_ini
```
_(the above create a ZeroLengthNew element with zero friction, with normal stiffness and tangential stiffness equal to 1e5)_
```
parameter 1 element 1 phi
updateParameter 1 [expr 16.7]
```
_(the above change friction coefficient in ZeroLengthNew element 1 from 0 to 0.3)_

*****************************
Detailed description of bugs in zeroLengthInterface2D:

Consider the following case, zeroLengthInterface2D or ZeroLengthNew is used in between a quad element and a series of beam elements. For both elements, the normal contact stiffness (Kn) and tangential stiffness (Kt) are 1e6 lbf/in, the friction coefficient is 0.3:
image

First, let the quad element and the beam elements have penetration (initiate normal contact pressure):
image

Second, let the quad element has cyclic sliding in horizontal direction:
image

Plot the friction vs horizontal displacement (delta)
1. For zeroLengthInterface2D:
image
2. For ZeroLengthNew
image

The normal contact force across the interface:
1. For zeroLengthInterface2D: 20000 lbf in step 1, 15000 lbf in step 2
2. For ZeroLengthNew: 10000 lbf in step 1, 10000 lbf in step 2.
[In step 1 (initiate the normal contact pressure), a normal contact force will be developed between node pair 1,8 and 4,7. At each node pair, the contact force is _Kn*0.005=1e6*0.005=5000 lbf_. Therefore, the correct normal contact force across the interface is 10000 lbf.]

### DIFF
--- a/SRC/element/TclElementCommands.cpp
+++ b/SRC/element/TclElementCommands.cpp
@@ -91,6 +91,7 @@ extern  void *OPS_ElasticTubularJoint(void);
 extern void *OPS_ZeroLengthContactNTS2D(void);
 extern void *OPS_ZeroLengthVG_HG(void);
 extern void *OPS_ZeroLengthInterface2D(void);
+extern void *OPS_ZeroLengthNew(void);
 extern "C" void *OPS_PY_Macro2D(void);
 extern void *OPS_SimpleContact2D(void);
 extern void *OPS_SimpleContact3D(void);
@@ -491,7 +492,14 @@ TclModelBuilderElementCommand(ClientData clientData, Tcl_Interp *interp,
       opserr << "TclElementCommand -- unable to create element of type : " << argv[1] << endln;
       return TCL_ERROR;
     }
-
+  } else if (strcmp(argv[1], "ZeroLengthNew") == 0) {
+      Element* theEle = (Element*)OPS_ZeroLengthNew();
+      if (theEle != 0)
+          theElement = theEle;
+      else {
+          opserr << "TclElementCommand -- unable to create element of type : " << argv[1] << endln;
+          return TCL_ERROR;
+      }
 
   } else if (strcmp(argv[1],"componentElement2d") == 0 || strcmp(argv[1],"componentElement") == 0) {
     void *theEle = 0;
--- a/SRC/element/zeroLength/ZeroLengthNew.cpp
+++ b/SRC/element/zeroLength/ZeroLengthNew.cpp
@@ -1,0 +1,838 @@
+/* ****************************************************************** **
+**    OpenSees - Open System for Earthquake Engineering Simulation    **
+**          Pacific Earthquake Engineering Research Center            **
+**                                                                    **
+**                                                                    **
+** (C) Copyright 1999, The Regents of the University of California    **
+** All Rights Reserved.                                               **
+**                                                                    **
+** Commercial use of this program without express permission of the   **
+** University of California, Berkeley, is strictly prohibited.  See   **
+** file 'COPYRIGHT'  in main directory for information on usage and   **
+** redistribution,  and for a DISCLAIMER OF ALL WARRANTIES.           **
+**                                                                    **
+** Developed by:                                                      **
+**   Frank McKenna (fmckenna@ce.berkeley.edu)                         **
+**   Gregory L. Fenves (fenves@ce.berkeley.edu)                       **
+**   Filip C. Filippou (filippou@ce.berkeley.edu)                     **
+**                                                                    **
+** ****************************************************************** */
+
+// $Source: /usr/local/cvs/OpenSees/SRC/element/zeroLength/ZeroLengthNew.cpp,v $
+// $Revision: 1.0 $
+
+// Written: Xianjue Deng  (UC San Diego, xianjue.deng@gmail.com)
+//
+// Date: March 02 2025
+
+/*
+ - element zeroLengthNew eleTag? -sNdNum sNdNum? -pNdNum pNdNum? –dof sdof? mdof? -Nodes Nodes? Kn? Kt? phi?
+ Description: This file contains the implementation for the ZeroLengthNew class.
+*/
+
+#include "ZeroLengthNew.h"
+#include <Information.h>
+#define PI 3.141592653589793238462643383279502884197169399
+#include <Domain.h>
+#include <Node.h>
+#include <Channel.h>
+#include <FEM_ObjectBroker.h>
+#include <Renderer.h>
+#include <Parameter.h>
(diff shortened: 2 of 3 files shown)


mhscott:
Yeah, the name needs to change to something more indicative of the element functionality.
You should also add documentation and the Python command.

XianjueDeng:
> I would recommend a different name, "New" is a bit generic and it looks like yours is an update to the zeroLengthInterface2D, so I would do a variant on that, maybe add the new feature in the name.

> Yeah, the name needs to change to something more indicative of the element functionality. You should also add documentation and the Python command.

Hi, I will change the element name accordingly. I am new here and I will figure out how to add the documentation and Python shortly. Thank you for the quick comments!

silviamazzoni:
It looks like your modified the original work by others (zeroLengthInterface2D.cpp) and removed the original-author names.
You need to keep the original authorship and add that a description of what you modified (or fixed) and added features.

the original element file, zeroLengthInterface2D.cpp contains the following lines, which you need to keep:
// Written: Roozbeh Geraili Mikola  (roozbehg@berkeley.edu)
//          Prof. Nicholas Sitar (nsitar@ce.berkeley.edu)
//
// Date: July 02 2010
add your info here....

silviamazzoni:
I think the element name should also maintain the original-element name since you have modified it. a different name will only cause confusion.

XianjueDeng:
> I think the element name should also maintain the original-element name since you have modified it. a different name will only cause confusion.

Hi, do you want me to have this new element the same name as zeroLengthInterface2D?

XianjueDeng:
> It looks like your modified the original work by others (zeroLengthInterface2D.cpp) and removed the original-author names. You need to keep the original authorship and add that a description of what you modified (or fixed) and added features.
> 
> the original element file, zeroLengthInterface2D.cpp contains the following lines, which you need to keep: // Written: Roozbeh Geraili Mikola ([roozbehg@berkeley.edu](mailto:roozbehg@berkeley.edu)) // Prof. Nicholas Sitar ([nsitar@ce.berkeley.edu](mailto:nsitar@ce.berkeley.edu)) // // Date: July 02 2010 add your info here....

Hi, I have placed this information in both the cpp file and at the beginning of the pull request. Thank you for this info.

silviamazzoni:
please also deal with the .h file.
and everything else that maintains the proper IP thread.

XianjueDeng:
> please also deal with the .h file. and everything else that maintains the proper IP thread.

Hi, I also added the information in the head file recognizing the original element. For the element name, if I include the full original name, will it be too long? For example, "zeroLengthInterface2DUpdated" or "zeroLengthInterface2DRevised"?

silviamazzoni:
please use a more sensical name than "update", unless you want to add update1, as there will likely be more updates. and update provides no insight into it. 
i would recommend you add something about the added feature.
or maybe just add your initials, as others have done
Usability is important!

XianjueDeng:
> please use a more sensical name than "update", unless you want to add update1, as there will likely be more updates. and update provides no insight into it. i would recommend you add something about the added feature. or maybe just add your initials, as others have done Usability is important!

Hi, thank you very much for the guidance. I used "update" because the new element updates the original elements by fixing bugs and added a new few feature that allow user to UPDATE friction coefficient during the analysis. So I think "update" contains two meanings and it is a good word. However, I totally agree with your suggestion. Maybe I call it "zeroLengthInterface2DXD"? (XD is my initial) Do you think it is a good choice?

mhscott:
@XianjueDeng If you find bugs in the original element, fix those too!

silviamazzoni:
> @XianjueDeng If you find bugs in the original element, fix those too!

thanks Mike, I had assumed that was a given.....

XianjueDeng:
> @XianjueDeng If you find bugs in the original element, fix those too!

Hi, my new element fixed all known bugs in the original element and added a new feature. I want to name the new element "zeroLengthInterface2DUpdated" (because it updates the original element by fixing bugs and the new feature is allowing user to UPDATE friction coefficient during the analysis) or "zeroLengthInterface2DXD" (XD is my initial). Could you please let me know which name looks good to you and what else I need to do?

mhscott:
I mean fix the bugs in the original element too. Don't just leave the bugs there.

XianjueDeng:
> I mean fix the bugs in the original element too. Don't just leave the bugs there.

Maybe I should submit this as a new version of the original element instead of a completely new element? In order to fix the bugs, I made significantly changes to the code of orignal element.

XianjueDeng:
> Yeah, the name needs to change to something more indicative of the element functionality. You should also add documentation and the Python command.

Hi, I have submitted a new PR in the OpenSeesDocumentation for adding the documentation. However, I am not sure how to add Python command. Can you please let me know how to do that?

mhscott:
Take a look at how other zero length elements are added in SRC/interpreter/OpenSeesElementCommands.cpp

PR #1585, opened 2025-03-12
adding PMLVISCOUS version 

amnp95:
Here is the PML Viscous implementation, an enhanced version of the previous PML implementation.

### DIFF
--- a/SRC/Makefile
+++ b/SRC/Makefile
@@ -683,6 +683,7 @@ ELE_LIBS   =  $(FE)/element/Element.o \
 	$(FE)/element/UWelements/Tcl_generateInterfacePoints.o \
 	$(FE)/element/PML/pml_3d.o \
     $(FE)/element/PML/PML3D.o \
+	$(FE)/element/PML/PML3DVISCOUS.o \
     $(FE)/element/PML/pml_2d.o \
     $(FE)/element/PML/PML2D.o \
 	$(FE)/element/PML/PML2D_3.o \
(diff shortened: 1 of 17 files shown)


amnp95:
@fmckenna
Hi Frank,

I made some updates to the PML implementation and also submitted a pull request for the documentation. However, I’m see in the pull request some compilation errors with OPS_GetNDMaterial and OPS_GetNodeCrd. I compiled in my ubuntu and stampede3 and tested the code but seems there is problem for the windows and some test for the OpenSeespySP . I attempted to troubleshoot and fix the issue but wasn’t able to resolve it.

Would you be able to take a look at the code when you have a chance? I’d really appreciate your insights.

Thank you!

fmckenna:
the commands exist in the tcl interpreter biut not in the python interpreter.
1. change OPS_GetNDMaterial to OPS_getNDMaterial
2. nothing exists that is similar to switch to. i will have to add somthing.

mhscott:
Why not get the node coordinates directly using a Node pointer instead of doing this circumnavigation?

image

mhscott:
This code is unnecessarily complicated.

+ Ask the user for an NDMaterial
+ Cast the NDMaterial to ElasticIsotropic
+ Get E, nu, rho parameters, then do some stuff (which should be done in the constructor btw)

Why not just take E, nu, and rho as element inputs? What happens when a user passes PDMY or something to this element?

image

PR #1591, opened 2025-03-25
Update CMakeLists.txt for mac

ma-wenqian:
The location of zlib and libaec have changed in the latest version of macOS. To address this, use the find_package command to atuomatically set the dir.


mhscott:
Same update in https://github.com/OpenSees/OpenSees/pull/1592

PR #1637, opened 2025-07-15
Fix memory leak in ElementStateParameter

jaabell:
Hi,

This commit fixes a memory leak in `ElementStateParameter`. There is a `fromFree` variable that is supposed to handle the case when memory ownership is external, but it is always set to 1 (which disables freeing memory in the destructor); however, the class always owns its memory. I decided to remove the `fromFree` variable altogether and maintain the behavior of always owning the memory. 

JAA


jaabell:
I also added the VTKHDF to makefile.s

PR #1647, opened 2025-08-05
Implement loadPackage command for the Python interpreter

aschellenberg74:
I was not sure which new additions to the matrix, vector, and versor classes should also be added to the DEVELOPER/core directory. Could you please look into that? Thanks!


mhscott:
There's still a lot superfluous changes in this PR that don't appear to be related to `loadPackage`.

Please update or close and make a new PR with a branch that is focused only on `loadPackage`

PR #1667, opened 2025-10-27
adding equivalent linear material

parduino:


mhscott:
OK, please fix it though. The compile is broken.

parduino:
Hi Frank,

  Just finished my last meeting of the day...
I implemented an equivalent linear constitutive model.
For this I added two files and modified several ones.
In the morning I went too into a pull request fast and when adding files I only committed the modified ones and not the two new files... This is why it choked.
I just created another pull request adding the two new files.
I thought you were reviewing all pull request before accepting them... If not I would have gone slower.
Sorry for the inconvenience.

Pedro

PR #1718, opened 2026-03-22
Fix index out of bounds in HystericSMMaterial

parched:
In particular, when pinch parameters weren't specified, index 0 of an empty vector was accessed.

### DIFF
--- a/SRC/material/uniaxial/HystereticSMMaterial.cpp
+++ b/SRC/material/uniaxial/HystereticSMMaterial.cpp
@@ -363,14 +363,14 @@ OPS_HystereticSMMaterial(void)
 
 
     // Parsing was successful, allocate the material
-    Vector theposEnv(&posEnv[0], (int)posEnv.size());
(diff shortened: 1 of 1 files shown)


silviamazzoni:
I think the change is a good improvement.

Replacing &vec[0] with vec.data() is the right fix for a possible empty-std::vector case. In this PR, that change is applied consistently to posEnv, negEnv, pinchArray, damageArray, degEnvArray, rotYArray, and also the two limit-state vectors. The previous code used &forceLimitStates[0] and &defoLimitStates[0] only when size was nonzero, but the first six vectors were still using &vec[0] directly. The new version removes that inconsistency and makes the construction cleaner.

From a C++ standpoint, data() is the safer interface here because it is specifically meant to expose the underlying contiguous buffer, whereas &vec[0] is problematic when the vector is empty. So as a bug fix, the PR makes sense.

I'll merge and test it at my end. Thanks @parched, glad the material is being used. Make sure you look at the fixes I made to the degradations in my last update.

silviamazzoni (approved):
looks good

PR #1735, opened 2026-04-11
Minor fixes for thermal analysis

johnnyontheweb:
- corrected wrong usage of delete for pointer in SteelECThermal, that was causing execution error
- more robust data handle in ForceBeamColumn3dThermal


mhscott:
Please update the PR to only modify source files, not the windows project files or compiler settings

PR #1739, opened 2026-04-18
Align system command parsing between OpenSeesTcl and OpenSeesPy.

gaaraujo:
Recently added options for linear solvers were not parsed correctly by the Tcl interpreter.


mhscott:
@gaaraujo I've been lazy about this and not set a good example :) but can you make the Tcl/commands.cpp call the OPS_Umfpack function so that we don't have to maintain two parsing functions moving forward? It's on our to-do list for many commands!

gaaraujo:
@mhscott Done! Did it for UmfPack and a few others, mostly for the serial versions of the solvers. Now they should be all using the same OPS_* interface.

PR #1742, opened 2026-04-23
Update CMakeLists.txt

mhscott:
Adding mumps dependency only if MUMPS_DIR is defined


gaaraujo:
Hi Michael! After this change I'm getting “dependency target mumps does not exist” when I pass `-DMUMPS_DIR=...`. Is that what you had in mind?

PR #1747, opened 2026-04-28
Fix null deref in TwoNodeLink

parched:
Depending on the response ID, theVector or theID may be null, so should only be dereferenced the response ID is checked.

Similar pattern was applied to a few other files to be safe, but for all current response IDs theVector happens to be non-null

Picked up when run with undefined behavior sanitizer


mhscott:
@parched We should check if the pointers are null before dereferencing, heh? Wasn't done before, but we can do it now.

parched:
We could do, but how far do you take it? As I understand it (but I'm fairly unfamiliar with the code), it should be impossible for these to be null where they're now dereferenced because the setResponse makes them non-null for these response IDs.

Does OpenSees have any convention for assertions, in particular, non-null assertions?

mhscott:
@parched Can you modify the code to do an early return if the point is null? The pointers should not be null, but it's a good idea to check any way.

```cpp
if (theVec == 0)
  return -1;
```

PR #1748, opened 2026-05-01
Newton solvers: add -factorOnce and consolidate Tcl/OPS parsers

gaaraujo:
This PR aims to improve a bit the efficiency of the Newton family of algorithms. Many files are touched, but the changes are conceptually small: it is mostly reorganization and parsing around how these solvers are built and how -factorOnce is interpreted.

Changes include:

1. Add -factorOnce support across the Newton family (Newton, Modified, Express, Accelerated, NewtonLineSearch). After the stiffness is formed and factored once, later Newton iterations in the same equilibrium solve reuse that factorization; reuse can continue across subsequent solves until something triggers a full domainChanged() path (for example nodes/elements added or removed, or anything that changes the equation system so the old factors are invalid). Then we form and factor again so the linear system size and storage stay consistent.
2. Accelerated Newton: -factorOnce is tied to forming the increment tangent once. If the user’s combination of -increment and -iterate flags would be inconsistent with that, we warn and default to either -iterate noTangent or disabling -factorOnce, depending on the case.
3. If -initial is used as the increment/iterate tangent in the supported Newton-style paths, factorOnce is turned on automatically.
4. Consolidate Tcl/Python algorithm construction in OpenSeesCommands and tcl/commands so both paths call the same OPS_ factories.
5, Attach the default convergence test to the algorithm during analysis setup so the algorithm and the interpreter’s default test stay wired together consistently (you can set the algorithm and still rely on the usual default test behavior).
6. Fixed bug in SecantAccelerator3: the secant update used vStar after it had already been overwritten, so the inner products / differences were taken against the wrong vector.

### DIFF
--- a/SRC/analysis/algorithm/equiSolnAlgo/AcceleratedNewton.cpp
+++ b/SRC/analysis/algorithm/equiSolnAlgo/AcceleratedNewton.cpp
@@ -46,12 +46,236 @@
 #include <Vector.h>
 #include <ID.h>
 
+#include <KrylovAccelerator.h>
+#include <RaphsonAccelerator.h>
+#include <SecantAccelerator1.h>
+#include <SecantAccelerator2.h>
+#include <SecantAccelerator3.h>
+#include <PeriodicAccelerator.h>
+
+#include <elementAPI.h>
+#include <string.h>
+#include <stdlib.h>
 #include <fstream>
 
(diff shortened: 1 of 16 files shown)


fmckenna:
@gaaraujo Gustavo .. can you update the documentation.

gaaraujo:
Done! https://github.com/OpenSees/OpenSeesDocumentation/pull/188

PR #1759, opened 2026-06-03
Adding error message when Invert() fails on section stiffness

mhscott:


mhscott:
@fmckenna We should change the signature of SectionForceDeformation::getSectionFlexibility so there's a return value that the calling objects can work with. Or we get rid of this function and let the calling objects do the inversion.

```cpp
int
SectionForceDeformation::getSectionFlexibility(Matrix &fs)
{
   const Matrix &ks = this->getSectionTangent();
   int ok = ks.Invert(fs);
   if (ok < 0)
     return ok;
   else
      return 0;
}
```

PR #1760, opened 2026-06-04
Adding G and Av as optional inputs for elastic beam (2d and 3d)

mhscott:
+ Optional additional inputs G and Av for 2d, Avy and Avz for 3d
+ Also read shear properties if section is the input for elastic constants


silviamazzoni:
FINALLY!!!!! thanks Mike!

PR #1762, opened 2026-06-05
Umfpack reuse numeric

gaaraujo:
Repeatedly computing the numeric factorization made Umfpack much slower than necessary when used with Modified or Krylov Newton methods.


mhscott:
@gaaraujo Awesome! Should help speed up `eigen()` when using UmfPack too

PR #1769, opened 2026-07-15
Expose METIS 5 mesh-partition controls in OpenSeesMP

masaqif:
# Expose METIS 5 mesh-partition controls in OpenSeesMP

## Summary

This change connects OpenSeesMP's Tcl `partition` command to the existing
METIS mesh-partition implementation and exposes commonly used METIS 5 tuning
controls. It also fixes the existing `-ufactor` parser bug and makes the MPI
build usable when MUMPS is not installed.

## Command syntax

```tcl
partition \
  -objective cut|volume \
  -ncuts number \
  -niter number \
  -ufactor number \
  -seed number \
  -info
```

Unknown options and invalid objectives now produce Tcl errors. `-info` reports
the selected objective, objective value, number of parts, and tuning options.

## Partition correctness

- Retain element loads only on the rank that owns their element.
- Keep nodal mass on one owning rank instead of duplicating it on interface
  node copies.
- Retain floating fixed nodes, their SP constraints, and their nodal loads.
- Align ownership for `equalDOF`/MP-constrained nodes and keep each constraint
  complete wherever its mesh node is retained.

These fixes and the associated state regression were adapted from #1773 with
Gustavo A. Araújo R. credited as co-author.

## Build changes

- Require system METIS 5 for the MPI mesh-partition command.
- Enforce the requested METIS major version in `FindMETIS.cmake`.
- Include the selected system METIS 5 header explicitly, avoiding collisions
  with OpenSees' bundled METIS 4 headers on Linux.
- Compile the shared interpreter with the partition feature when MPI is found.
- Link OpenSeesMP to the selected METIS and Tcl libraries.
- Pass MPI C++ compile options as a list so multi-option MPICH configurations
  are handled correctly.
- Compile MUMPS sources only when a MUMPS installation is configured.
- Allow users and toolchains to override the macOS deployment target.
- Let CMake select compatible Fortran runtimes instead of injecting static GCC
  runtimes into every C++ target.
- Avoid exporting a `stat` data symbol from DistributedSuperLU, which can
  override the Linux C library's `stat()` function at runtime.
- Warn from rank zero when unsupported `EQ_Constraint` objects are present.\n
## Benchmark

The included benchmark constructs a 1,660-element 2D truss grid on every rank
and verifies conservation after partitioning.

| Ranks | Objective | METIS value | Elements per rank | Maximum imbalance |
|---:|---|---:|---|---:|
| 2 | volume | 42 | 832, 828 | 0.24% |
| 4 | volume | 112 | 410, 417, 416, 417 | 0.48% |
| 8 | volume | 186 | 209, 211, 208, 211, 211, 202, 211, 197 | 1.69% |
| 4 | cut | 65 | 408, 411, 421, 420 | 1.45% |

Repeating the four-rank volume test with seed 42 produces identical objective
and per-rank element counts.

## Verification

```text
macOS 26 / MPICH 5.0.1        OpenSeesMP build passed; pytest: 7 passed
Ubuntu 24.04 / Open MPI 4.1.6 OpenSeesMP build passed; pytest: 7 passed
Ubuntu 24.04 / MPICH 4.2.0    OpenSeesMP build passed; pytest: 7 passed
```

The Ubuntu MPICH-linked binary was run with a PMIx-aware launcher because the
ARM64 distribution's Hydra/PMIx combination created singleton communicators
even for a minimal MPI smoke test. `ldd` confirmed that OpenSeesMP loaded
MPICH, not Open MPI.

The pytest suite covers 2-, 4-, and 8-rank conservation and balance, fixed-seed
determinism, both METIS objectives, invalid-objective diagnostics, clean MPI
shutdown, element-load ownership, single-owner nodal mass, floating fixed-node
retention, nodal loads, and mesh/floating `equalDOF` ownership.

### DIFF
--- a/SRC/actor/machineBroker/MPI_MachineBroker.cpp
+++ b/SRC/actor/machineBroker/MPI_MachineBroker.cpp
@@ -63,6 +63,7 @@ MPI_MachineBroker::~MPI_MachineBroker()
   delete [] theChannels;
   delete usedChannels;
 
+  MPI_Barrier(MPI_COMM_WORLD);
   MPI_Finalize();
 }
 
@@ -124,4 +125,3 @@ MPI_MachineBroker::freeProcess
   // channel not found!
   return -1;
 }
-
--- a/SRC/actor/machineBroker/MPI_MachineBroker.cpp
+++ b/SRC/actor/machineBroker/MPI_MachineBroker.cpp
@@ -63,7 +63,6 @@ MPI_MachineBroker::~MPI_MachineBroker()
   delete [] theChannels;
   delete usedChannels;
 
-  MPI_Barrier(MPI_COMM_WORLD);
   MPI_Finalize();
 }
 
--- a/SRC/interpreter/OpenSeesMiscCommands.cpp
+++ b/SRC/interpreter/OpenSeesMiscCommands.cpp
@@ -58,7 +58,7 @@
 #include <BackgroundMesh.h>
 #ifdef OPS_PARTITION_WITH_METIS
 #include <mpi.h>
-#include <metis.h>
+#include OPS_METIS_HEADER
 #endif
 
 #ifdef _OPENMP
--- a/SRC/interpreter/OpenSeesMiscCommands.cpp
+++ b/SRC/interpreter/OpenSeesMiscCommands.cpp
@@ -1992,7 +1992,7 @@ int OPS_partition() {
     int ufactor = 30;
     int seed = -1;
     int objective = METIS_OBJTYPE_VOL;
-    int info = 0;
+    bool printInfo = false;
     while (OPS_GetNumRemainingInputArgs() > 0) {
         int num = 1;
         auto opt = OPS_GetString();
@@ -2044,7 +2044,7 @@ int OPS_partition() {
                 return -1;
             }
         } else if (strcmp(opt, "-info") == 0) {
-            info = METIS_DBG_INFO;
+            printInfo = true;
         } else {
             opserr << "WARNING: unknown partition option " << opt << "\n";
             return -1;
@@ -2154,7 +2154,6 @@ int OPS_partition() {
         options[METIS_OPTION_NUMBERING] = 0;
         options[METIS_OPTION_UFACTOR] = ufactor;
         options[METIS_OPTION_SEED] = seed;
-        options[METIS_OPTION_DBGLVL] = info;
 
         // number of parts to partition
         idx_t nparts = np;
@@ -2180,7 +2179,7 @@ int OPS_partition() {
             return -1;
         }
 
-        if (info != 0) {
+        if (printInfo) {
             opserr << "METIS partition: objective="
                    << (objective == METIS_OBJTYPE_CUT ? "cut" : "volume")
                    << " value=" << objval
--- a/SRC/interpreter/OpenSeesMiscCommands.cpp
+++ b/SRC/interpreter/OpenSeesMiscCommands.cpp
@@ -56,7 +56,7 @@
 #include <TetMesh.h>
 #include <Damping.h>
 #include <BackgroundMesh.h>
-#ifdef _PARALLEL_INTERPRETERS
+#ifdef OPS_PARTITION_WITH_METIS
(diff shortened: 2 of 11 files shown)


gaaraujo:
@mhscott @fmckenna @zhuminjie

I had some bug fixes and additional ideas for `OPS_partition` (see #1773), but they would overlap with this PR, so I closed mine.

Happy to wait for @masaqif’s PR to be reviewed and merged, then open a follow-up PR on top if that works better. Just let me know how you’d like to proceed.

masaqif:
Thanks @gaaraujo — I incorporated the partition-correctness fixes from #1773 in commit 002be9d and credited you as co-author. The updated implementation now keeps element loads with their owning elements, assigns nodal mass to a single owner, retains floating fixed nodes and their loads/constraints, and aligns ownership for mesh-to-floating and floating-to-floating MP/equalDOF pairs.

I also added a focused four-rank regression covering all of those cases. The full seven-test MPI suite passes on macOS/MPICH, Ubuntu/Open MPI, and an Ubuntu MPICH-linked build.

The additional -samePart, -customPartition, and parallel ARPACK work are valuable new features, but I have kept them out of this PR to avoid expanding its review scope. A follow-up PR on top after this merges sounds ideal.

gaaraujo:
Thanks @masaqif! I put the parallel ARPACK stuff on PR #1774 since it's really independent from the partition stuff. As for the `-samePart` and `-customPartition`, I agree that they are better left for a follow up PR that I can take care of!

If it's okay with you, I'm going to leave some comments on your code for your consideration.

SRC/actor/machineBroker/MPI_MachineBroker.cpp
```diff
@@ -63,6 +63,7 @@ MPI_MachineBroker::~MPI_MachineBroker()
   delete [] theChannels;
   delete usedChannels;
 
+  MPI_Barrier(MPI_COMM_WORLD);
```
gaaraujo: I think `MPI_Finalize` is enough on its own. You do not need a manual MPI_Barrier before it. Let me know if I'm missing something!

SRC/interpreter/OpenSeesMiscCommands.cpp
```diff
@@ -2183,7 +2238,43 @@ int OPS_partition() {
         }
     }
 
-    // get nodes in current processor and remove mp
+    // Give nodes connected by an MP constraint consistent ownership on every
```
gaaraujo, line 2247: Could we issue a warning from rank 0 when `domain->getNumEQs() > 0`? `EQ_Constraint` (multiple retained nodes linked to a single constrained node) needs different handling than `MP_Constraint`, and `OPS_partition()` as is does not cover that case. Just so users know! A follow-up PR can fix that later.

SRC/actor/machineBroker/MPI_MachineBroker.cpp
masaqif: Agreed. I removed the explicit barrier in 93dc7ac and now rely on `MPI_Finalize` directly. I rebuilt OpenSeesMP, ran the seven-test MPI suite five consecutive times under MPICH on macOS, and reran it under Linux Open MPI and with an MPICH-linked Linux binary; all runs passed and exited cleanly.

SRC/interpreter/OpenSeesMiscCommands.cpp
masaqif: Added in 93dc7ac. `OPS_partition()` now emits a single rank-zero warning, including the number found, whenever `domain->getNumEQs() > 0`; the warning states that `EQ_Constraint` partitioning is not yet supported and results may be incorrect. The updated OpenSeesMP builds and the full MPI regression suite pass with MPICH and Open MPI.

masaqif:
Hi @fmckenna @mhscott — the outstanding review comments on #1769 have been addressed and both threads are resolved. The seven-test MPI suite passes under MPICH and Open MPI. The CMake workflow is still marked `action_required` and has not executed any jobs. When convenient, could a maintainer approve the workflow and take another look at the PR? Thanks!

PR #1773, opened 2026-07-21
Enable METIS 5 partitioning and parallel eigen analysis in OpenSeesMP

gaaraujo:
## Summary

This PR enables model partitioning in `OpenSeesMP` through the existing `OPS_partition` implementation already used by `OpenSeesPy`. It also enables parallel ARPACK eigenvalue analysis with `system Mumps` and `numberer ParallelPlain`.

## Changes

- Exposes the existing [`partition`](https://openseespydoc.readthedocs.io/en/latest/src/partition.html) command in `OpenSeesMP`.

- Adds METIS 5 discovery and linking for Linux and Windows. OpenSees includes a bundled copy of METIS 4 for legacy graph partitioning, so the build configuration keeps the METIS 4 and METIS 5 headers and libraries separate.

- Fixes several correctness issues in `OPS_partition`, including:
  - retaining element loads only on the rank that owns the corresponding element;
  - assigning nodal mass to a single owning rank;
  - retaining fixed and constrained floating nodes appropriately; and
  - maintaining consistent ownership for `equalDOF`/MP-constrained nodes.

- Adds `-samePart`, which constrains groups of elements to the same METIS partition. This is useful for elements such as springs using [`PyLiq1`](https://openseespydoc.readthedocs.io/en/latest/src/PyLiq1.html), which obtain state information from other elements.

  Each group is specified by its element count followed by the element tags:

  ```tcl
  partition -samePart 3 101 102 103 \
            -samePart 2 201 202
  ```

- Adds `-customPartition`, allowing users to provide explicit element-to-part assignments instead of using METIS. Part IDs are zero-based.

  ```tcl
  partition -customPartition 4 \
      1 0 \
      2 0 \
      3 1 \
      4 1
  ```

  Unlisted elements are assigned to part 0 with a warning.

- Enables parallel ARPACK eigenvalue analysis by providing `ArpackSOE` with the MPI process ID and communication channels already used by the parallel system of equations. Eigenvalue analysis can therefore be performed both before and after partitioning.

- Emits a warning when equation constraints are present because their partitioning behavior is not yet fully supported.

## Example

A self-contained 3D brick example is provided at:

```text
EXAMPLES/Partition/Example.tcl
```

It compares serial execution with standard METIS partitioning, custom partitioning, and `-samePart`. The example prints gravity displacements, eigenvalues, and transient displacements for direct comparison.

For example:

```bash
OpenSees Example.tcl serial 4 4 20
mpiexec -n 4 OpenSeesMP Example.tcl metis 4 4 20
mpiexec -n 4 OpenSeesMP Example.tcl custom 4 4 20
mpiexec -n 4 OpenSeesMP Example.tcl samePart 4 4 20
```

The example was verified with 2–8 MPI processes.

## Build notes

METIS 5 is readily available through Linux package managers, but setting it up on Windows requires building it separately. Installation instructions are provided in:

```text
README-Windows-METIS5.md
```

Example `OpenSeesMP` build scripts are also included:

```text
makeOpenSeesMP_Ubuntu.sh
makeOpenSeesMP_WIN.bat
```

These changes allow models that previously relied on `OpenSeesSP` partitioning to run in `OpenSeesMP` with minimal script changes. Although the primary focus is `OpenSeesMP`, the fixes within `OPS_partition` also benefit `OpenSeesPy` users.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,6 +121,12 @@ else()
     include(OpenSeesFunctions) # Your custom logic
 endif()
 
+# Conan MSVC packages ship Release import libs only; map other optimized configs.
+if(CMAKE_CONFIGURATION_TYPES)
+  set(CMAKE_MAP_IMPORTED_CONFIG_RELWITHDEBINFO Release)
+  set(CMAKE_MAP_IMPORTED_CONFIG_MINSIZEREL Release)
+endif()
+
 find_package(ZLIB REQUIRED)
 find_package(HDF5 REQUIRED)
 find_package(TCL REQUIRED)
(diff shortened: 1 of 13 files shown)


gaaraujo:
Closing, already covered by https://github.com/OpenSees/OpenSees/pull/1769 !

Issue #1779, opened 2026-07-31
Allow CatenaryCable to be used with ndof=6


apalazzi:
Currently it's not possible to use CatenaryCable elements with e.g. elasticBeamColum in a model with dof=6, but I need this feature for some of my calculations, so I added it: would you be interested in merging it in the current code?

I'll prepare a report of what I had to do to make it work and other things I found, together with the sources.

The results with the modified version matches the validation test of the current version, plus other tests; there was a discrepancy on weight and load appliation, I'll write about it with more detail.

Obviously I couldn't test the interaction with beam and other elements in the current version, if someone can give me some reference example to run I'll add them also.

PR #1788, opened 2026-08-06
Support CatenaryCable with six-DOF nodes

masaqif:
## Summary

CatenaryCable currently requires three-DOF nodes, preventing its use with six-DOF beam nodes. This change accepts matching three- or six-DOF nodes and maps the existing translational formulation into six- or twelve-entry element interfaces. Cable rotational stiffness, mass, and forces remain zero.

Element-sized matrices and vectors are instance-owned. Force recorder labels match the element DOF count, and the recorder's XML ElementOutput tag is closed so multiple elements and their data are correctly structured.

Addresses #1779.

## Verification

23 CatenaryCable cases cover:

- the documented OpenSeesWiki displacement benchmark for both node types;
- matching three-/six-DOF displacements and end forces;
- self-weight reactions and an independent symmetric elastic-catenary span relation;
- mixed three-/six-DOF cables in both construction orders;
- lumped, integration-fallback, and equivalent-truss mass assembly, with zero cable rotational mass;
- lumped and equivalent-truss inertia forces, including nonzero nodal angular accelerations;
- XML force labels, values, and sibling element structure;
- coupling to an elastic beam with free, nonzero nodal rotations.

The reference/equilibrium coverage was prompted by apalazzi's comparison with #1796. The benchmark source is the [OpenSeesWiki CatenaryCableElement example](https://opensees.berkeley.edu/wiki/index.php/CatenaryCableElement).

Local validation on macOS arm64 using Python 3.11:

```console
cmake --build build/regression --target OpenSeesPy -j4
/opt/homebrew/bin/python3.11 tests/run_pytest.py --module build/regression/OpenSeesPy.dylib --tests-dir tests
git diff --check
```

Result: build succeeded; all 30 tests passed, including the 23 CatenaryCable cases. The new runner loads the specified compiled extension directly. The expanded tests run in the existing Ubuntu test job; focused cable test steps are also included for macOS and Windows.

The branch includes upstream master through 2890cb36f, including the Conan layout and macOS Fortran library-path fixes relevant to the previous CI failures. The new [CMake Build run](https://github.com/OpenSees/OpenSees/actions/runs/34324353955) is marked `action_required` with no jobs started; remote CI validation is awaiting maintainer approval.

Contributor: Musawer Ahmad Saqif <saqif@umich.com>


apalazzi:
@masaqif I had an LLM do the same, I didn't submit it because I wanted to review it but haven't had the time.

The PR is here, if you want to take a look at it and see it: https://github.com/OpenSees/OpenSees/pull/1796

I had a LLM compare the two, here is the report:

PR #1788 (masaqif) vs Our PR

### PR #1788: "Support CatenaryCable with six-DOF nodes"
- **3 files, +154/−64**
- Author: `masaqif` (Musawer Ahmad Saqif, UMich)
- Addresses issue #1779
- Status: **Open**, not merged

|                          | PR #1788                                                  | Our PR                                                         |
|--------------------------|-----------------------------------------------------------|----------------------------------------------------------------|
| **Matrices**             | Instance variables (non-static), resized in `setDomain()` | Static class members + 12-DOF variants                         |
| **`numDOF` derivation**  | `numDOF = 2 * dofNd1`                                     | `if (dofNd1==3) numDOF=6; else if (dofNd1==6) numDOF=12`       |
| **DOF validation**       | `dofNd1 != dofNd2 \|\| (dofNd1 != 3 && dofNd1 != 6)`      | Separate if/else with warning                                  |
| **0-DOF fix**            | ❌ No                                                     | ✅ Yes (15 LinearSOE files)                                    |
| **Python packaging**     | ❌ No                                                     | ✅ `setup.py`, `pyproject.toml`, `MANIFEST.in`                 |
| **Tests**                | 2 simple (3-DOF, 6-DOF)                                   | 6 comprehensive (analytical, reference, self-weight × 3/6-DOF) |
| **Reference comparison** | ❌                                                        | ✅ JSON + generation script                                    |
| **`setResponse()`**      | Dynamic tagging (f1–f12)                                  | Still hardcoded f1–f6 ⚠️                                        |
| **`Mass.Zero()`**        | At top of `computeMass()`                                 | Per-method                                                     |
| **`load->Zero()`**       | ✅                                                        | ✅                                                             |
| **`disp.Zero()`**        | ❌                                                        | ✅                                                             |

### Key Design Differences

1. **Static vs Instance:** PR #1788 makes `Stiffness`/`Mass`/`ZeroMatrix`/`Forces` **instance** members (one per element). We keep them **static** (shared across all instances) + separate 12-DOF variants. Their approach is cleaner — static shared state is a bug risk if multiple elements have different DOF counts.

2. **`setResponse()`:** PR #1788 dynamically tags response names for 12-DOF. We still hardcode f1–f6. ⚠️ **We should fix this.**

3. **`Mass.Zero()`:** PR #1788 zeros `Mass` at the top of `computeMass()` (one place). We zero per-method.

masaqif:
@apalazzi Thank you for sharing #1796 and the comparison. I reviewed the implementation, test cases, and reference data and expanded the verification in 5be8b9f3f.

The cable suite now has 23 passing cases: the OpenSeesWiki displacement benchmark in both DOF configurations; matching translational displacements/end forces; self-weight equilibrium and an independent symmetric elastic-catenary span relation; mixed-DOF elements; mass and inertia-force mapping; XML recorder labels and values; and coupling to a beam with nonzero rotations. The test source credits your comparison for prompting the broader reference/equilibrium coverage.

The recorder checks also reproduced an existing missing `output.endTag()` in `CatenaryCable::setResponse()`. This is fixed, with tests checking that multiple ElementOutput records are siblings and Data is at the correct level.

Regarding `disp.Zero()`: `Vector(int)` already initializes every entry to zero, so the local `Vector disp(numDOF)` initializes the unused rotational entries. The mass/inertia and beam-coupling tests also check zero cable rotational contributions.

The zero-equation solver changes and Python packaging in #1796 are separate from the node-DOF extension. The fully restrained self-weight tests here use the element equilibrium established when it is added to the domain, without invoking a zero-equation linear solve. I also kept the existing `addLoad()` behavior: the implementation assigns the distributed load components, whereas the additional self-weight-plus-eleLoad tests in #1796 expect accumulation. Changing that behavior would need a separate compatibility decision.

I merged current upstream master (2890cb36f), which includes the Conan layout and macOS Fortran library-path fixes for the previous CI failures, and added cable test steps to macOS/Windows alongside the existing Ubuntu test job. The local OpenSeesPy build succeeds and the full tests directory passes 30/30.

The new [CMake Build run](https://github.com/OpenSees/OpenSees/actions/runs/34324353955) is marked `action_required`, with no jobs started. Could a maintainer approve the workflow run so the updated branch can receive remote CI validation?

PR #1789, opened 2026-08-06
Handle Matrix and Vector otherFact special cases

### LINKED ISSUE #1687: otherFact not checked for 0, -1, 1 in some Matrix and Vector algebra functions
@fmckenna Kinda a note for us to discuss next time we meet. For example, `Matrix::addMatrixProduct` doesn't check for the special cases of `otherFact`

masaqif:
## Summary

Fixes #1687 by handling `otherFact` values of `0`, `1`, and `-1` consistently in Matrix and Vector algebra helpers.

The zero-factor path now avoids reading operand storage and correctly applies `thisFact` to the destination. The `+1` and `-1` paths avoid unnecessary multiplication while preserving the existing algebra. The affected Matrix methods include matrix addition, transpose addition, matrix products, transpose products, and both triple-product overloads. The affected Vector methods include vector addition, matrix-vector products, and transpose matrix-vector products.

The patch also adds a focused CTest executable covering all affected methods and the special-factor behavior.

## Verification

- `cmake --build build/regression --target matrix-vector-otherfact-test -j4`
- `ctest --test-dir build/regression --output-on-failure -R matrix-vector-otherfact`
- Result: `1/1` tests passed.
- `git diff --check`

An additional `_G3DEBUG` build was attempted. It is currently blocked by the pre-existing `Vector::operator%` error at `SRC/matrix/Vector.cpp:1237`, where the `Matrix`-returning function returns `-1`.

Submitted by **Musawer Ahmad Saqif** <saqif@umich.com>.

### DIFF
--- a/SRC/matrix/Matrix.cpp
+++ b/SRC/matrix/Matrix.cpp
@@ -44,6 +44,16 @@ using std::nothrow;
 
 #include <math.h>
 
+static inline double
+scaleWithFact(double value, double fact)
+{
+  if (fact == 1.0)
+    return value;
+  if (fact == -1.0)
+    return -value;
+  return value * fact;
+}
+
 int Matrix::sizeDoubleWork = MATRIX_WORK_AREA;
 int Matrix::sizeIntWork = INT_WORK_AREA;
 double Matrix::MATRIX_NOT_VALID_ENTRY =0.0;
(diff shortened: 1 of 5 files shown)


SRC/matrix/Matrix.cpp
```diff
@@ -688,7 +707,7 @@ Matrix::addMatrix(double factThis, const Matrix &other, double factOther)
 	double *dataPtr = data;
 	double *otherDataPtr = other.data;		    
 	for (int i=0; i<dataSize; i++)
-	  *dataPtr++ = *otherDataPtr++ * factOther;
+	  *dataPtr++ = scaleWithFact(*otherDataPtr++, factOther);
```
gaaraujo, line 710: I think it would be preferable to check the value of factOther outside the loop, and then avoid the multiplication. Otherwise you're repeating that check in every step of the for loop. Same for the other parts of the code.

masaqif: Addressed in 0e9b13a1f: removed scaleWithFact and dispatch otherFact values 1, -1, and the general case before entering the Matrix loops. Vector now uses multiplication only in its existing general-factor branches. I also expanded the regression test to cover general otherFact and nontrivial thisFact values. OPS_Matrix builds successfully and the focused CTest passes 1/1. Thank you for the review.

PR #1791, opened 2026-08-10
ForceBeamColumn: fix partial uniform load Vz sign

parched:


mhscott:
@parched The sign is consistent with the local axes of the basic system in 3d and dM/dx=V instead of dM/dx=-V. I guess the issue is the sign of the internal shear force doesn't match the local z-axis direction?

https://openseesdigital.com/2022/09/18/the-basic-natural-kernel-in-a-corotating-frame-of-reference/

image

PR #1792, opened 2026-08-10
Fix memory leaks in SymmGeneralizedEigenSolver

parched:
On solve failure, the early return was leaking memory


fmckenna:
fixed with this pull request (preferred to stick to our normal older c++ in mem allocation) https://github.com/OpenSees/OpenSees/commit/01315fb6e609e101fd8d1b0176496febd723d47d

PR #1793, opened 2026-08-11
Account for shear stiffness in elastic beam column

parched:
Also fix/implement some other loading cases.

This was written entirely by Claude, however, it does now match the results of some hand calcs that led to this fix and further independent tests written by Claude.

> ElasticBeam2d/3d: shear-consistent fixed-end forces and end releases
> 
> These elements have always used a Timoshenko stiffness but Euler-Bernoulli
> fixed-end forces, so element loads produced end moments inconsistent with the
> stiffness resisting them. Map the fixed-end moments through
> M_T = k_T k_EB^-1 M_EB, which is exact for an arbitrary load distribution
> because shear leaves the simply supported primary end rotations alone.
> 
> Releases were wrong for the same reason: the branches hard-coded the
> Euler-Bernoulli carry-over of 1/2 rather than (2-phi)/(4+phi). This changes
> results even for a full-span uniform load, whose fixed-end moments are
> otherwise shear-invariant.
> 
> Also implement release handling that was simply missing:
>   - 2D partial uniform loads ignored -release silently
>   - 3D partial uniform loads ignored -releasey/-releasez (warned once)
>   - 3D point loads had no release handling at all
> 
> Factor the shared algebra into ElasticBeamCommon.h, replacing five
> copy-pasted correction blocks and nine ad-hoc release branches.
> 
> Elements defined without a shear area give phi = 0 and are unaffected.

### DIFF
--- a/SRC/element/elasticBeamColumn/CMakeLists.txt
+++ b/SRC/element/elasticBeamColumn/CMakeLists.txt
@@ -8,6 +8,7 @@ target_sources(OPS_Element
     PRIVATE
       ElasticBeam2d.cpp
       ElasticBeam3d.cpp
+      ElasticBeamCommon.h
       ElasticBeamWarping3d.cpp      
       ElasticTimoshenkoBeam2d.cpp
       ElasticTimoshenkoBeam3d.cpp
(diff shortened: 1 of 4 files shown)


mhscott:
Thanks!

PR #1802, opened 2026-09-06
Hysteretic, HystereticSM: restore the tangent in revertToLastCommit(); SuperLU: set Fact before sp_preorder() in setSize()

angshuman311:
Two problems found running a displacement-controlled pushover of a wall building (MVLEM_3D with HystereticSM shear springs) with SuperLU and an adaptive step size.

1. Hysteretic and HystereticSM did not restore the tangent in revertToLastCommit(). After a failed step the integrator forms the tangent before any new setTrialStrain() is called, so every retry of the step used the tangent from the failed trial. When that trial had reached a flat segment of the envelope the tangent was zero, dgstrf() reported a singular matrix, and reducing the step size could not recover. Added Ctangent, set in commitState() and restored in revertToLastCommit(), as Concrete01 does. It is also set in revertToStart(), copied in getCopy() and sent in place of Ttangent in sendSelf(). Converged results are unchanged.

2. SuperLU::setSize() called get_perm_c() and sp_preorder() with options.Fact still set to SamePattern from the previous factorization and set it back to DOFACT only afterwards. sp_preorder() computes the column etree only when Fact is DOFACT. So on any domainChanged() during an analysis (new integrator, algorithm or test, or a model change between analyze() calls) dgstrf() ran with the old etree and a new perm_c. On Windows this crashes with heap corruption in dgstrs(); elsewhere it can produce wrong factors. Fact is now set before sp_preorder(). setSize() also frees the previous L, U, AC, A, B and stat, which were leaked. The GlobalLU_t passed to dgstrf() was an uninitialized local in solve(); dgstrf() reads it when Fact is SamePattern_SameRowPerm (the -symm option), so it is now a zeroed member of the solver.


mhscott:
Please separate these changes into two PRs so that we can review separately the Hysteretic and SuperLU changes

angshuman311:
Split into #1808 (Hysteretic, HystereticSM) and #1809 (SuperLU) as requested. Closing this one.

PR #1811, opened 2026-09-12
Fix the SSPquadUP fluid storage at zero solid density and its lower-side pressure load

fkkarakurt:
`SSPquadUP` drops the pore fluid's storage and stabilization terms whenever the solid phase has no mass, and sends half of its lower-side pressure to a pore pressure DOF instead of a vertical one. Both show up in a one-element model. I found them while checking the u-p elements against Terzaghi and Mandel consolidation; there is no open issue for either.

### What goes wrong

One element, `E = 1e4`, `nu = 0.3`, `Kf = 2e4`, `e = 2/3` (so `n = 0.4`), unit thickness, `rho = 0`:

| Case | Exact | Before |
|---|---|---|
| Sealed element under `q = 10`, undrained pore pressure | 7.8788 | 10.0 |
| Column of 20 elements drained at the top, base pressure at `T = 0.1` | 7.4794 | 9.0896 |
| `Plow = 10` on a unit square: vertical reaction at node 2 | 5.0 | 0.0 |
| the same run: pore pressure reaction at node 1 | 0.0 | 5.0 |
| `Pup`, `Pleft` and `Pright` | exact | exact |

```python
import openseespy.opensees as ops

E, nu, Kf, n, perm, q = 1.0e4, 0.3, 2.0e4, 0.4, 1.0e-6, 10.0
e = n/(1 - n)
M = E*(1 - nu)/((1 + nu)*(1 - 2*nu))

# 1. sealed element on a massless solid: undrained pore pressure
ops.model('basic','-ndm',2,'-ndf',3)
for tag,(x,y) in enumerate(((0,0),(1,0),(1,1),(0,1)),1):
    ops.node(tag,float(x),float(y))
ops.fix(1,1,1,0); ops.fix(2,1,1,0); ops.fix(3,1,0,0); ops.fix(4,1,0,0)
for tag in (2,3,4):
    ops.equalDOF(1,tag,3)
ops.nDMaterial('ElasticIsotropic',1,E,nu,0.0)
ops.element('SSPquadUP',1,1,2,3,4,1,1.0,Kf,1.0,perm,perm,e,1.0e-6)
ops.timeSeries('Constant',1); ops.pattern('Plain',1,1)
ops.load(3,0.0,-q/2,0.0); ops.load(4,0.0,-q/2,0.0)
ops.constraints('Transformation'); ops.numberer('RCM'); ops.system('BandGeneral')
ops.test('NormDispIncr',1.0e-12,10,0); ops.algorithm('Linear')
ops.integrator('Newmark',0.5,0.25); ops.analysis('Transient')
ops.analyze(1,0.1)
print(ops.nodeVel(1,3), q/(1 + n*M/Kf))   # 10.0, exact 7.8788

# 2. a pressure of 10 on the lower side of a unit square
ops.wipe()
ops.model('basic','-ndm',2,'-ndf',3)
for tag,(x,y) in enumerate(((0,0),(1,0),(1,1),(0,1)),1):
    ops.node(tag,float(x),float(y)); ops.fix(tag,1,1,1)
ops.remove('sp',3,1)   # one free DOF off the loaded side, so the system is not empty
ops.nDMaterial('ElasticIsotropic',1,E,nu,0.0)
ops.element('SSPquadUP',1,1,2,3,4,1,1.0,Kf,1.0,perm,perm,e,0.0,0.0,0.0,0.0,10.0,0.0,0.0)
ops.timeSeries('Constant',1); ops.pattern('Plain',1,1)
ops.constraints('Plain'); ops.numberer('Plain'); ops.system('FullGeneral')
ops.test('NormDispIncr',1.0e-12,10,0); ops.algorithm('Linear')
ops.integrator('LoadControl',1.0); ops.analysis('Static')
ops.analyze(1); ops.reactions()
for tag in (1,2):
    print(ops.nodeReaction(tag,2), ops.nodeReaction(tag,3))
# node 1: 5.0 5.0 before, 5.0 0.0 after
# node 2: 0.0 0.0 before, 5.0 0.0 after
```

### Why

**The storage terms.** `getMass()` assembles the element's `[ M 0 ; 0 -S ]`, where the pressure block carries the compressibility `n/Kf` and the pressure stabilization `Kp`. It computed both, then returned the zeroed matrix if the solid density was zero, before either reached the matrix. A massless solid is the ordinary choice for a quasi-static consolidation run, and it left the analysis with an incompressible pore fluid and no stabilization. That is why the sealed element above answers `q` rather than `q/(1 + nM/Kf)`.

The 3D sibling `SSPbrickUP` assembles the same block regardless of the solid density, and drops the solid mass in `GetSolidMass` instead. `SSPquadUP::GetSolidMass` already scales every term by the density, so removing the early return leaves the solid block at zero on its own and only restores the fluid block.

**The lower side.** `setPressureLoadAtNodes()` fills a 12-vector ordered `ux1 uy1 p1 ux2 uy2 p2 ...`. The contribution of the lower side puts its horizontal components at 0 and 3, which is right, and its vertical components at 1 and 2. Index 2 is the pore pressure DOF of node 1; the vertical DOF of node 2 is 4. So half the traction never reached the solid, and entered the fluid mass balance as a spurious flux instead. The other three sides use the right indices, which is why only `Plow` is affected.

### Change

- `getMass()`: the early return is gone, with a comment on why the fluid block has to survive a massless solid.
- `setPressureLoadAtNodes()`: `pressureLoad(2)` becomes `pressureLoad(4)`.

### Tests

`tests/test_ssp-quad-up.py` covers both, on a massless solid throughout:

- the undrained pressure of a sealed element, `q/(1 + nM/Kf)`, which Newmark with `gamma = 1/2` reproduces exactly at any step size, together with the matching volumetric strain
- a 20-element column against Terzaghi's series with the diffusivity `perm/(n/Kf + 1/M)` of a compressible fluid, in pressure along the column and in surface settlement
- each of `Pup`, `Plow`, `Pleft` and `Pright` on a unit square, requiring `P/2` along the outward normal at both nodes of the loaded side and nothing on any pore pressure DOF

Before the change the undrained, consolidation and `Plow` cases fail and the other three sides pass; after it, all six pass.

Built and tested on Ubuntu with the same CMake/conan configuration as the CI workflow. The full `tests/` suite passes (108 tests).

Documentation companion: OpenSees/OpenSeesDocumentation#197


fkkarakurt:
The Windows job fails in the MUMPS step, before anything in this PR is compiled.

All three sources listed for `MUMPS_5.5.1.tar.gz` in `OpenSees/mumps` (`cmake/libraries.json`) are currently unreachable. From here:

```
http://opensees.berkeley.edu/OpenSees/code/MUMPS_5.5.1.tar.gz   no response
http://mumps.enseeiht.fr/MUMPS_5.5.1.tar.gz                     no response
https://graal.ens-lyon.fr/MUMPS/MUMPS_5.5.1.tar.gz              404
```

`FetchContent` then fails at configure time and the job ends with `ninja: error: loading 'build.ninja'`. Ubuntu and macOS pass.

The same workflow passed on Windows for #1810 on 11 September at 19:50 UTC, on the same base commit (`01315fb6e`), so this is not from the change here, and every Windows run will fail until a reachable source is restored.

`https://mumps-solver.org/MUMPS_5.5.1.tar.gz` serves the same archive. Its sha256 is `1abff294fa47ee4cfd50dfd5c595942b72ebfcedce08142a75a99ab35014fa15`, which is the checksum already recorded for 5.5.1, so adding it to that version's `urls` list should be enough.

Issue #1812, opened 2026-09-15
Anomalous PWP/effective stresses at imposed-pressure nodes in OpenSees 3.8 (works in 3.7.1), u-p dynamic analysis


mariglenhutka-hub:
Hi everyone,

I ran into something puzzling and I'm hoping someone here can help me understand whether it's a bug or something I'm doing wrong.
I'm working on nonlinear dynamic effective stress analysis of an earth dam under earthquake shaking, as part of a seismic fragility study. The model is a 2D coupled solid-fluid (u-p) mesh, so every node has 3 DOFs (ux, uy and pore water pressure). What I care about is how pore water pressure and effective stress evolve in the dam and foundation during earthquake, especially along the submerged upstream face and the reservoir bed.

The analysis has two stages: first gravity stage (elastic and plastic) to set up the initial stresses under the reservoir, then a dynamic stage where I apply the ground motion.

Along the submerged upstream boundary, I prescribe the hydrostatic pore pressure at each node using a `MultipleSupport` pattern with `imposedMotion` on the pressure DOF, and I also apply the reservoir water loads at those nodes:
```
timeSeries Constant 3000 -factor 361.008
pattern MultipleSupport 3000 {
    groundMotion 2999 Plain -vel 3000
    imposedMotion 12457 3 2999
    # ... one line per upstream node, grouped by prescribed pressure value
}
```

Here is the strange part. I ran the exact same model, same mesh, same boundary conditions, same loads, same solver settings, in two versions of OpenSees:

- In 3.7.1, everything works well. The pore pressures and effective stresses look physically reasonable everywhere, including the upstream face, through both the gravity and dynamic stages.
- In 3.8, the gravity stage is still fine, but the moment the dynamic stage begins, the pore pressures and effective stresses at those upstream prescribed - pore water pressure nodes (the reservoir bed and the upstream slope) blow up to very large, unphysical values.

When I narrow the contour range in post-processing. I can see that the values at the interior of the mesh are fine, it's only the upstream surface goes wrong.

What makes it more confusing is that the analysis converges without complaint in 3.8. I don't get any convergence errors, the solver reports normal convergence at every step, it just produces these wrong values at the boundary. I'll attach some of the convergence log so you can see it's converging cleanly.

```
step 391 complete
plastic gravity phase complete
damping coefficients: a_0 = 0.04976780441980199;  a_1 = 0.00031515830311111683
Permeabilities Updated
Sides x-dir fixities removed
load pattern applied
[...]

PR #1557, opened 2024-12-18
Derive theAccelSeries from theDispSeries in GroundMotion

yulee:
Addressing Issue #1555

PR #1813, opened 2026-09-15
Fix acceleration derivation scope for ground motions

silviamazzoni:
I implemented a fix . However, I was able to only run unit tests, but need the real modellers to verify the fix.

## Issue

OpenSees 3.8 introduced a regression affecting imposed motions defined using only a velocity time series. This was reported in issue #1812 for a u–p dynamic analysis where pore-water pressure is prescribed using a constant time series and a MultipleSupport pattern.

For u–p elements, the third nodal degree of freedom represents pore pressure. OpenSees stores the pressure value in the node’s velocity vector and its time derivative in the acceleration vector. A constant hydrostatic pressure should therefore produce:

p = 361.008  
dp/dt = 0.0

In OpenSees 3.7.1, a GroundMotion supplied with only a velocity series used that series directly and left the corresponding acceleration at zero.

PR #1557 added support for deriving missing ground-motion series. As part of that change, GroundMotion began automatically differentiating a supplied velocity series to generate acceleration. This behavior was added globally, even though it was principally needed by UniformExcitation, which forms its excitation load from acceleration.

The differentiation algorithm assumes an initial preceding value of zero and uses the default GroundMotion time increment of 0.01. Differentiating the constant prescribed pressure therefore generates an artificial first acceleration value:

2 × (361.008 − 0) / 0.01 = 72,201.6

A minimized reproduction confirmed that the imposed node received:

velocity = 361.008  
acceleration = 72,201.6

ImposedMotionSP places these values directly into the node’s trial velocity and acceleration vectors. For SSPquadUP, the artificial DOF-3 acceleration is then consumed by getResistingForceIncInertia() as a pressure-rate contribution. This produces the anomalous pore-water pressures and effective stresses reported in issue #1812.

The underlying problem was therefore not specific to SSPquadUP or to pore-pressure degrees of freedom. Derivation required by one consumer, UniformExcitation, had been introduced as default behavior for every GroundMotion consumer, including MultipleSupport and ImposedMotionSP.

## Summary of PR

Restore pre-3.8 behavior for velocity-only MultipleSupport ground motions while preserving velocity- and displacement-derived acceleration for UniformExcitation.

## Root cause

PR #1557 enabled missing-series differentiation globally in GroundMotion. Consequently, a constant velocity used by ImposedMotionSP was differentiated with an initial zero value, creating a large artificial acceleration. For u-p models, this acceleration occupies the pore-pressure-rate DOF and enters element inertia calculations.

## Fix

- Restore non-deriving GroundMotion behavior for missing acceleration and velocity.
- Add an explicit virtual `deriveAccelSeries()` operation.
- Invoke it from UniformExcitation, the consumer that requires acceleration.
- Propagate the opt-in through InterpolatedGroundMotion.

No Tcl or OpenSeesPy syntax changes are required.

## Tests

- MultipleSupport constant velocity preserves pressure and produces zero pressure rate.
- UniformExcitation derives acceleration from velocity.
- UniformExcitation derives acceleration from displacement.
- Explicit acceleration behavior is unchanged.
- Existing BandGeneral and force-beam-column smoke tests pass.

Fixes Issue #1812


silviamazzoni:
@parduino have a look!

